PR #3, opened 2016-08-12
Remove protocol prefix

tiagojsag:


nerik:
Ideally, LiveReload should actually be disabled when not running locally.

nerik:
Fixed by https://github.com/Vizzuality/TRASE-frontend/pull/4

PR #10, opened 2016-08-28
Project architecture

nerik:


src/index.js
```diff
@@ -0,0 +1 @@
+console.log('hello');
```
tiagojsag: ups?

src/reducers/app.reducer.js
```diff
@@ -0,0 +1,14 @@
+import a from 'actionNames';
```
tiagojsag: 0 points for creativity on variable names

davidsingal:
So, finally we are going to use Redux. I agreed. But I think we have to standardise the way to create projects, could you check this project? https://github.com/resource-watch/prep-app

nerik:
@davidsingal I don't know exactly what you mean by standardise: 
- naming conventions: definitely agree, but as long as there are no formalized guidelines...
- file tree: same
- code style: one .eslintrc to rule them all, y hasta
- toolchain: webpack is pretty much the default choice, then there has to be some custom config depending on the project
- using a starter template: it's a case by case thing, in many cases I think they tend to be too opinionated 
- libraries/frameworks: case by case again. Why do you point me to a React project?

Let's have a chat ;)

clementprdhomme:
There's a lot of work going out here. I really like webpack's aliases!

As Tiago pointed out, there are some small changes to do. And I would like to add some more:
- the license of the project should be MIT (can you confirm @davidsingal?)
- you don't have `autoprefixer` configured (don't forget the `.browserslist`, ask me for details or see other projects)

About the architecture, I'm totally fine with Redux and this project would be a good test for it. Nevertheless, as David said, the structure of the folders is completely different from the other projects and I'm personally confused with it. I would prefer to keep with what we've done in the past to make sure everyone could at some point join the project without being lost.

In addition, I would really appreciate if you could comment the code. For someone (like me) who hasn't been implicated in it, it's really hard to understand what every function does.

nerik:
@davidsingal @clementprdhomme ok, it's time to have formalized Vizzuality standards then. Do we have anything like this? If not I'll go ahead and create a repo.

simaob:
Hi @nerik There's been an attempt of having some guidelines and style guide which has been added to the intranet. It would be great to have comments on there and to improve and then make sure everyone is aware of them! =)

nerik:
Squashed and hardcore-force-pushed to develop

nerik:
Hi @simaob. We discussed about this with @davidsingal and @clementprdhomme, there's a project guidelines repo on the works.

PR #12, opened 2016-09-05
Basic styling & CSS modularity

agnlez:
This PR splits all current styles placed in `flows` in separated components. it also defines clearly the `flows` layout.
- Added a reset CSS.
- Added some class helpers and mixins.
- Navigation menu looks quite better (not finished)

There's a file called `node.scss` in `components/sankey` defines all elements inside the sankey (node, links, labels..). Not sure if we should divide this file in smaller ones in case styles start to grow more.


.sass-lint.yml
```diff
@@ -0,0 +1,17 @@
...
+    - 2
+    -
+      allow-rgba: true
+  no-url-protocols:
+    - 0
+  nesting-depth:
+    - 2
+    -
+      max-depth: 3
+  no-vendor-prefixes: 0
```
nerik, line 17: I think we should enforce this, since we have autoprefixer in the project.

nerik, line 17: Should we set everything to warnings (1) instead of errors (2) ?
If we add sass linting as a webpack build step, it probably never should break for sass errors. What do you think?

styles/_base.scss
```diff
@@ -0,0 +1,33 @@
+@import 'reset';
+@import 'settings';
```
nerik, line 2: You don't need this import.

styles/_mixins.scss
```diff
@@ -0,0 +1,31 @@
+@mixin triangle($size, $border-width, $color, $direction) {
```
nerik, line 1: Nice 👍

styles/components/map.scss
```diff
@@ -0,0 +1,7 @@
+
+.c-map {
```
nerik, line 2: Component global layout (position, size) is not internal to the component, and should not be defined at component level, but at page layout level.

styles/components/map/map-legend.scss
```diff
@@ -0,0 +1,10 @@
+@import '../../settings';
+
+.c-map-legend {
```
nerik, line 3: Same as https://github.com/Vizzuality/TRASE-frontend/pull/12/files#r77595822

styles/components/nav.scss
```diff
@@ -0,0 +1,16 @@
+@import '../settings';
+
+.c-nav {
+  display: flex;
+  width: 100%;
+  height: 56px;
```
nerik: Same as https://github.com/Vizzuality/TRASE-frontend/pull/12/files#r77595822

styles/components/sankey/node.scss
```diff
@@ -1,41 +1,42 @@
...
+  opacity: .2;
+  stroke: $medium-pink;
 }
+
 .sankey-state-links .sankey-link {
-  opacity: 0.2;
-  stroke: black;
+  opacity: .2;
+  stroke: $black;
 }
```
nerik: This file is named node.scss but defines style for all elements of the sankey (background, nodes, links). Why not symply naming it sankey.scss ?

styles/components/timeline.scss
```diff
@@ -0,0 +1,6 @@
+
+.c-timeline {
```
nerik: Same as https://github.com/Vizzuality/TRASE-frontend/pull/12/files#r77595822

styles/_base.scss
agnlez: Why? You are using it to set font weights in this case.

styles/components/map.scss
agnlez: Agree.

styles/_base.scss
nerik: Sorry, I've read too fast :/

styles/components/sankey/node.scss
agnlez: Well, I thought about it but I also thought it was kind of redundant but if you are OK, no problem!

nerik:
sankey/sankey.scss and sankey.scss should be a single file IMHO.
Otherwise, good to merge 👍

agnlez:
Merging...!

PR #13, opened 2016-09-06
Add git hooks

tiagojsag:
Add git hooks


tiagojsag:
@nerik can you check if we can add the js lint status - can we enforce the validation before commiting?

nerik:
I'll have a look today, FYI jslint is run at transpile time by webpack (fatal errors prevent webpack to complete)

nerik:
Looks good. The hooks are a bit complicated to test, but this has been battle-tested in fishing right?

PR #14, opened 2016-09-06
Feature/resize by

nerik:
This enables the user to reload the sankey with a different kind of value encoded in links ('resize by').
Right now we are rebuilding the whole diagram. in the near future, we should be able to simply redraw, opening the possibility of smooth transitions.


agnlez:
looks fine! :)

PR #17, opened 2016-09-07
Menu Styles & Bookmark menu implementation

agnlez:
- Added Handlebars & Handlebars-loader (don't forget run `npm install`)
- Added new SVG icons
- Added new componentes views for bookmark menu and bookmark items.
- Added helper class to parse string to HTML.
- Styles for nav and bookmark menu.


nerik:
Now we have templating capabilities, cool 👍 
My only remark here: there is no need to have different templating languages for templates at webpack level (ejs) and at runtime level (handlebars). I don´t have any preference, so you should pick the one you prefer - if we go with handlebars, lets replace ejs files and webpack plugin with handlebars equivalents.

config/webpack.js
```diff
@@ -20,13 +20,15 @@ const pages = {
 
 const htmlHeadTemplate = _.template(fs.readFileSync('./html/includes/_head.ejs', 'utf8'));
 const htmlFooterTemplate = _.template(fs.readFileSync('./html/includes/_footer.ejs', 'utf8'));
+const htmlIconsTemplate = _.template(fs.readFileSync('./html/includes/_icons.ejs', 'utf8'));
```
nerik: This is a static file, no need to import it as a template. Just have a .svg file and import the raw text as is in the template.

html/flows.ejs
```diff
@@ -8,6 +8,13 @@
   <div class="l-flows">
 
     <div id="nav" class="c-nav is-absolute">
+      <div class="menu">
+        <div class="burger">
+          <span class="burger-ingredient"></span>
```
nerik: ROFL 😆

scripts/components/nav/bookmarks/bookmark-item.component.js
```diff
@@ -0,0 +1,47 @@
...
+  constructor(settings) {
+    this.el= settings.el;
+    this.bookmark = settings.bookmark;
+    this.actions = ['update', 'share-bookmark', 'delete'];
+
+    return this;
+  }
+
+  _setEventListener() {
+    this.bookmarkItem.querySelector('.update').addEventListener('click', () => this._update());
```
nerik: What do you think about introducing jQuery at some point.

scripts/templates/bookmark.handlebars
```diff
@@ -0,0 +1,17 @@
+<li
```
nerik: File name: I think the convention is .hbs file extension now? I haven´t tested but its mostly about syntax highlighting in Atom/whatever IDE

scripts/templates/bookmark.handlebars
```diff
@@ -0,0 +1,17 @@
...
+  class="bookmark-item"
+  data-bookmark-id="{{ bookmark.id }}"
+  data-bookmark-state="{{ bookmark.state }}"
+>
+  {{ bookmark.name }}
+  <ul class="bookmark-options">
+    {{#each actions}}
+      <li class="bookmark-action {{ this }}">
+        <svg class="icon icon-{{ this }}">
+          <use xlink:href="#icon-{{ this }}"></use>
```
nerik: That´s cool.

scripts/components/nav/bookmarks/bookmark-item.component.js
agnlez: Well, it's true it makes thing easier but I thought you didn't want to implement it because of some reason (usually performance)  so.. I dealt with it.  Actually, I don't have any preference so.. It's up to you :D

scripts/templates/bookmark.handlebars
agnlez: Yes you can go this `.hbs` also. Although As we are going to use EJS this is pointless ^^!

config/webpack.js
agnlez: I think `.svg` is not going to work  (the file contains HTML hides SVG definitions). I just moved as `html` file to a `statics` folder, is it OK for you? We still import it as we did.

agnlez:
I think we can go ahead with EJS. I setted Handlebars because I didn't know about EJS functionality at all and I need it. But seems works fine for our goals so let's go ahead.

config/webpack.js
nerik: You don´t need to call _.template I think, just drop the string result of readFileSync

html/index.ejs
```diff
@@ -4,13 +4,7 @@
   <%= htmlWebpackPlugin.options.head %>
 </head>
 <body>
-  <p>
-    Welcome to TRASE :)
-  </p>
-
-  <p>
-    <a href='./flows.html'>Bring me to the sankey page</a>
```
nerik, line 12: If you remove this @tiagojsag is gonna be very unhappy.

scripts/components/nav/bookmarks/bookmark-item.component.js
```diff
@@ -0,0 +1,47 @@
...
+export default class {
+  constructor(settings) {
+    this.el= settings.el;
+    this.bookmark = settings.bookmark;
+    this.actions = ['update', 'share-bookmark', 'delete'];
+
+    return this;
+  }
+
+  _setEventListener() {
```
nerik: setEventListeners

scripts/components/nav/bookmarks/bookmark-item.component.js
nerik: We´ll see if the need arises I guess :)

nerik:
No worries. I also made a small change in the webpack config for icons, if that looks good to you we can merge :)

PR #22, opened 2016-09-20
New api

nerik:


tiagojsag:
there's a DS_Store file commited here, otherwise merge

PR #25, opened 2016-09-20
Loading component

nerik:
Adds a simple loading bit that could serve for the map, etc, too.
Also move the sankey SVG to a container and cleans up some stuff.


agnlez:
I just modify some CSS in order to center the component properly, check it out @nerik please. If you agree, merge!

nerik:
Perfect, thanks!

PR #26, opened 2016-09-22
Map vector layers

nerik:
screen shot 2016-09-26 at 11 40 54

Shows 3 types of geo entities at the same time for now, eventually they will be toggled by changing the 1st column.


agnlez:
looks fine!

PR #29, opened 2016-09-26
Columns selectors

nerik:
Adds a dropdown to select the displayed column under each column.
Shows municips/states/biomes on the map depending on the 1st column.
Reload get_links and updates sankey accordingly.
It has not been extensively tested against the API considering the unstable state of it.


agnlez:
Great job!! We have to work a little more selector's styles although I think we can work on it later.

PR #31, opened 2016-09-27
Feature/basemap selector

nerik:


scripts/components/map-layers.component.js
```diff
@@ -0,0 +1,169 @@
...
+
+export default class {
+
+  onCreated() {
+    this._setVars();
+    this._setEventListeners();
+  }
+
+  selectedVectorLayers(layers) {
+    this.state = {
```
nerik: This is a violation of the 'unique source of truth' idea (basically do not duplicate information in the component internal state while it is already in the store). I don't think you even need this `state` obect, since you get the information from the store anyway and it's just passed around inside the component.
LMK if it's not clear I'll try to code it.

scripts/containers/map-layers.container.js
```diff
@@ -0,0 +1,15 @@
+import connect from 'connect';
+import { selectVectorLayers, selectContextualLayers } from 'actions/flows.actions';
+import mapLayers from 'components/map-layers.component';
+
+const mapMethodsToState = (state) => ({
+  selectedVectorLayers: state.flows.selectedVectorLayers,
```
nerik, line 6: Just `selectVectorLayers` ?

scripts/containers/map-layers.container.js
```diff
@@ -0,0 +1,15 @@
...
+  selectContextualLayers: state.flows.selectedContextualLayers
+});
+
+const mapViewCallbacksToActions = () => ({
+  onSelectedVectorLayers: layers => selectVectorLayers(layers),
+  onSelectedContextualLayer: layers => selectContextualLayers(layers)
```
nerik: `onVectorLayersSelected`, `onContextualLayerSelected`

scripts/actions/flows.actions.js
```diff
@@ -34,6 +34,14 @@ export function selectColumn(columnIndex, columnId) {
   };
 }
 
+export function selectVectorLayers(vectorLayers, reloadLinks) {
+  return _getSelectAction('vectorLayers', vectorLayers, actions.SELECT_VECTOR_LAYERS, reloadLinks);
```
nerik: When you toggle a vector layer, I'm not exactly sure where the data is gonna come from, but it'll be either:
- using data from the previous getNodes call
- or using a distinct API call
- or calling Carto SQL API

This data will not come from the get_links endpoint, so we should not use `_getSelectAction`. Just dispatch a  `SELECT_VECTOR_LAYERS` or `SELECT_CONTEXTUAL_LAYERS` action directly and we'll figure out later what should happen.

scripts/actions/flows.actions.js
```diff
@@ -34,6 +34,14 @@ export function selectColumn(columnIndex, columnId) {
...
+}
+
+export function selectContextualLayers(contextualLayers, reloadLinks) {
+  return _getSelectAction('contextualLayers', contextualLayers, actions.SELECT_CONTEXTUAL_LAYERS, reloadLinks);
```
nerik: `_getSelectAction` was a wrong choice of name, can you rename it `_reloadLinks` or something like this plz?

nerik (commented):
This looks really good, just a few changes.

nerik:
Also: the whole sankey should be pushed to the right, not only the map.

nerik:
Also: the layer selector should also be triggered when clicking on the map legend.

scripts/components/map.component.js
```diff
@@ -64,4 +70,9 @@ export default class {
     });
     return topoLayer;
   }
+
+  _onToggleBasemap() {
+    document.querySelector('.c-basemap-options').classList.toggle('-open');
+    document.querySelector('.c-map').classList.toggle('-moved');
```
nerik: It would be ideal here if instead you would dispatch an action, and make the "layer selector" visible/not-visible part of the store.

scripts/containers/map-layers.container.js
agnlez: Pardon? I don't understand this. I also have `selectContextualLayer` method.

scripts/containers/map-layers.container.js
agnlez: My bad! >,<

agnlez:
`Also: the layer selector should also be triggered when clicking on the map legend.`

This is something I would change in the map legend PR.

PR #35, opened 2016-10-06
Select node links from map / select map polygons from sankey

nerik:
trase

Additionally, adds a feature that keeps the node selection +_links whan changing years/indicator etc


agnlez:
aweeesome job! Merging..

PR #39, opened 2016-10-10
Feature/toggle map

agnlez:
- Added toggle map function.
- Added several animations when layout moves.
- Fixed some issue coming from develop
- Changes to keep linters happy


nerik:
Nice. I'm merging but there will be a few changes needed:
- [ ] as per design links should be hidden in full screen map mode (https://app.zeplin.io/project.html#pid=57cd3cdee754aa1a7bd6cbba&sid=57f612efaba6fb602aa9ee37)

screen shot 2016-10-10 at 15 34 01
- [ ] expand button should overlap map
  screen shot 2016-10-10 at 15 33 06
- [ ] map should go back to its center point when collapsing (now it's in the middle of the Atlantic)
- [ ] trigger an invalidate on the map before expanding to avoid the map blinking
- [ ] overall transitions are too long

PR #48, opened 2016-10-17
Feature/factsheets page

nerik:


nerik:
A few comments:
- we don't need redux, containers and actions for this portion of the site. This will just get in our way - the page is mostly non interactive and there is no need to store any state in the URL.
- we are finishing writing the spec for the places and actors fact sheets pages. You can check the full spec by browsing here: https://traseplatform.org/api/v1/ui/
- please try to keep PRs as small as possible so we can have less WIP PRs and more merged stuff

scripts/pages/factsheets.page.js
```diff
@@ -0,0 +1,24 @@
+import 'styles/factsheets.scss';
+
+import AreaStack from 'scripts/statics/factsheets/area-stack';
+
+class factsheetsIndex {
```
nerik: You don't need a class here. Just have the code call some functions and make classes for code that actually need encapsulation.

scripts/pages/index.page.js
```diff
@@ -0,0 +1,8 @@
+
+import 'styles/homepage.scss';
+
+class homepageIndex {
```
nerik: You don't need a class here. Just have the code call some functions and make classes for code that actually need encapsulation.

styles/_settings.scss
```diff
@@ -38,9 +43,13 @@ $font-weight-bold: 700;
 $font-size-small: 10px;
 $font-size-regular: 12px;
 $font-size-x-regular: 13px;
+$font-size-xx-regular: 16px;
```
nerik, line 46: I think we have too many font sizes in the project. Please check Zeplin or with JC.

public/factsheets/sample.json
```diff
@@ -0,0 +1,20 @@
+{
+  "metadata": {
+    "includedYears": [2014, 2015, 2016]
+  },
+  "data": [
```
nerik, line 5: Note: the API is similar but not exactly the same. Please have a look at the swagger UI for more details.

scripts/pages/factsheets.page.js
```diff
@@ -0,0 +1,24 @@
+import 'styles/factsheets.scss';
+
+import AreaStack from 'scripts/statics/factsheets/area-stack';
```
nerik: What does `/statics/` mean ?
Area-stack should simply go to ´components´ IMHO

scripts/pages/factsheets.page.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  _renderAreaStack() {
+    new AreaStack({
+      el: this.municipalitiesGraph
+    });
+  }
+
+  _setVars() {
+    this.el = document.querySelector('.l-factsheets');
```
nerik: Please use class names prefixed with `js-` when using querySelector.

scripts/statics/factsheets/area-stack.js
```diff
@@ -0,0 +1,151 @@
...
+  stack as d3_stack
+} from 'd3-shape';
+
+import {
+  axisBottom as d3_axis_bottom,
+  axisLeft as d3_axis_left
+} from 'd3-axis';
+
+import { json as d3_json } from 'd3-request';
+import { extent as d3_extent } from 'd3-array';
```
nerik: You managed to find all D3 imports. Congrats, you earned this GIF:

scripts/statics/factsheets/area-stack.js
```diff
@@ -0,0 +1,151 @@
...
+    const height = this.options.height - margins.top - margins.bottom;
+
+    const svg = d3_select(this.el).append('svg')
+      .attr('width', width + margins.left + margins.right)
+      .attr('height', height + margins.top + margins.bottom);
+
+    // scales
+    const x = d3_scale_time().range([0, width]);
+    const y = d3_scale_linear().range([height, 0]);
+    const z = d3_scale_ordinal(d3_schemeCategory_10);
```
nerik: This will need to change at some point if we want to use preset colors.

scripts/statics/factsheets/area-stack.js
```diff
@@ -0,0 +1,151 @@
...
+        .attr('class', 'area')
+        .style('fill', function(d) { return z(d.key); })
+        .attr('d', area);
+
+      layer.filter(function(d) { return d[d.length - 1][1] - d[d.length - 1][0] > 0.01; })
+        .append('text')
+          .attr('x', width - 6)
+          .attr('y', function(d) { return y((d[d.length - 1][0] + d[d.length - 1][1]) / 2); })
+          .attr('dy', '.35em')
+          .style('font', '10px sans-serif')
```
nerik: Move to CSS.

styles/_reset.scss
```diff
@@ -83,3 +83,11 @@ h1, h2, h3, h4, h5, h6 {
   margin: 0;
   padding: 0;
 }
+
+input {
+  -ms-appareance: none;
+  -moz-appearance: none;
+  -webkit-appearance: none;
```
nerik: Aren't we autoprefixing? I'm not sure, could you check?

nerik (commented):
A few comments, otherwise looks good!

styles/_settings.scss
agnlez: Yeah.. I am waiting for the UI Kit to discard sizes/colours... But since we don't have it yet I don't want to mess it up :/

scripts/statics/factsheets/area-stack.js
agnlez: This was, literally, a copy-paste from the example  😂

styles/_reset.scss
agnlez: We are. My bad.

scripts/statics/factsheets/area-stack.js
agnlez: I'll move custom colours to an array in `constants` file.

scripts/statics/factsheets/area-stack.js
agnlez: x

nerik:
👍

PR #53, opened 2016-10-18
API-driven map

nerik:
- renders the map layers list from an API call
- renders the choropleth from an API call

The map is basically not working because the data is not ready on the API side yet, but this is ready to review. @andresgnlez


scripts/reducers/sankey/getChoropleth.js
```diff
@@ -0,0 +1,56 @@
+import _ from 'lodash';
+
+export default function(selectedVectorLayers, nodesDictWithMeta, allColors) {
+  const horizontalLayer = selectedVectorLayers.horizontal;
+  const verticalLayer = selectedVectorLayers.vertical;
+  const isBidimensional = horizontalLayer.uid && verticalLayer.uid ? true : false;
```
agnlez: I think this can be reduced to `horizontalLayer.uid && verticalLayer.uid`

scripts/reducers/sankey/getChoropleth.js
```diff
@@ -0,0 +1,56 @@
...
+  const choropleth = {};
+
+  nodeIds.forEach(nodeId => {
+    const node = geoNodes[nodeId];
+    let color = 'none';
+
+    if (isEmpty) {
+      color = 'none';
+    } else if (!node.meta) {
+      color = '#0f0';
```
agnlez: I think literal colours should be moved to constants, this way we control theme easily and they are located in one place.

agnlez:
Just a fewer comments, otherwise looks great! Merge it once you read/apply them

Issue #18, opened 2016-09-08
Add API.fetch polyfill


agnlez:
http://caniuse.com/#search=fetch

App is not working on IE11 and Safari 9 at this moment.

nerik:
Yes.
(We might want to move this issues in PT).

PR #54, opened 2016-10-20
Fix/cross browser

nerik:
Fixes https://github.com/Vizzuality/TRASE-frontend/issues/18

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,8 @@
     "pretty-error": "^2.0.0",
     "redux": "^3.5.2",
     "redux-thunk": "^2.1.0",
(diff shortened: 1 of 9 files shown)


nerik:
@andresgnlez There are a few CSS issues with Safari and FF on the top menu and some icons, could you please have a look?

agnlez:
There's an [issue](https://bugzilla.mozilla.org/show_bug.cgi?id=874718) with Firefox and the radios in the map-layers component, and it's displaying wrongly.. We have to take a look at this later.

nerik:
Thanks @andresgnlez !

PR #56, opened 2016-10-21
AND node selection

nerik:
This changes node selection rules. Instead of selecting all links that pass through any of the selected nodes, links are selected if at each column, they pass through one of the selected nodes for this column or the column does not contain selected nodes.
Basically this is an AND node selection instead of OR.


nerik:
(also fixes a bug that caused top bar node information to not match the actual node selection order)

agnlez:
`Basically this is an AND node selection instead of OR.` <- perfect recap. Works fine! I don't know if you want someone else to check the PR, otherwise I think it's ready to merge.

PR #57, opened 2016-10-22
Node hover

nerik:
Adds hover effect and cross-selection between sankey and map.


scripts/containers/titlebar.container.js
```diff
@@ -7,7 +7,12 @@ const mapMethodsToState = (state) => ({
   selectNodes: state.flows.selectedNodesData,
   highlightNode: {
     _comparedValue: (state) => state.flows.highlightedNodeData,
-    _returnedValue: (state) => (state.flows.highlightedNodeData.length === 0) ? state.flows.selectedNodesData : state.flows.highlightedNodeData
+    _returnedValue: (state) => {
+      if (state.flows.highlightedNodeData.length === 0) {
```
agnlez, line 11: isn't this the same as `!state.flows.highlightedNodeData.length` ? ( 0 == falsy value)

scripts/components/titlebar.component.js
```diff
@@ -4,21 +4,21 @@ export default class {
...
+    this._toggle(selectedNodesData.length > 0);
   }
 
-  _toggle(nodesData) {
-    if (nodesData.length === 0) {
-      this.search.classList.remove('is-hidden');
-      this.nodesTitles.classList.add('is-hidden');
-    } else {
+  _toggle(showTitles) {
+    if (showTitles === true) {
```
agnlez: isn't `(showTitles)` enough?

scripts/components/titlebar.component.js
```diff
@@ -3,13 +3,22 @@ export default class {
...
+    this._toggle(selectedNodesData.length > 0);
+  }
+
+  _toggle(showTitles) {
+    if (showTitles === true) {
       this.search.classList.add('is-hidden');
       this.nodesTitles.classList.remove('is-hidden');
+    } else {
+      this.search.classList.remove('is-hidden');
+      this.nodesTitles.classList.add('is-hidden');
```
agnlez: I think you can use conditional `toggle` here, no?

agnlez:
Should't the app updates when you click on a polygon in the map? Or this functionality is still not implemented?

nerik:
Depends if the corresponding node exists in the left column of the sankey (most of the time it doesnt)
Yes, we need to explain visually that the node is disabled.

agnlez:
Looks good!

scripts/components/titlebar.component.js
nerik: Yup

PR #58, opened 2016-10-22
link transitions

nerik:
D3 ✨ 

transition


agnlez:
woo! Now sankey feels like the real deal!

PR #59, opened 2016-10-24
Feature/homepage page

agnlez:


scripts/pages/index.page.js
```diff
@@ -1,2 +1,46 @@
 
+import Dropdown from 'scripts/components/dropdown.component';
+
 import 'styles/homepage.scss';
+
+const defaults = {
```
nerik: mapDefaults ?

scripts/pages/index.page.js
```diff
@@ -1,2 +1,46 @@
...
+  const commodity = defaults.commodity.toLowerCase();
+  const country = defaults.country.toLowerCase();
+  const imageName = `${commodity}-${country}`;
+
+  imagesItem.forEach((imageItem) => {
+    if (imageItem.getAttribute('data-image-name') === imageName) {
+      imageItem.classList.remove('is-hidden');
+    } else {
+      if (!imageItem.classList.contains('is-hidden')) {
+        imageItem.classList.add('is-hidden');
```
nerik: You can use

```
classList.toggle('is-hidden', imageItem.getAttribute('data-image-name') === imageName)
```

styles/components/homepage/workflow-graphic.scss
```diff
@@ -0,0 +1,77 @@
...
+      flex-basis: 20%;
+      flex-direction: column;
+      align-items: center;
+      justify-content: center;
+
+      z-index: 1;
+
+      &:first-child > .icon-supply,
+      &:nth-child(4) > .icon-supply,
+      &:last-child > .icon-supply {
```
nerik: Please use the more stable `icon-supply_chain_4` class names instead, ie if we change order or add items, it will not break the CSS.

html/index.ejs
```diff
@@ -19,34 +19,80 @@
...
-            the trading of
-            <select name="commodity" id="">Soy</select>
+        <div class="c-map-risks">
+          <h3 class="title -big -center -light">What are the sustainability risks associated with
+            the trading <br/>of
+            <div class="c-dropdown -homepage" data-dropdown="commodity">
+              <span class="js-dropdown-title dropdown-title">
+                -
+              </span>
+              <ul class="js-dropdown-list dropdown-list">
```
nerik, line 63: Dropdowns overlap text which seems a bit weird. They should either show up below, or take the same space and be visually similar as the title text (check with JC)

screen shot 2016-10-24 at 12 36 49

nerik (changes requested):

scripts/pages/index.page.js
agnlez: You don't toggle the class here, it only displays selected item and hide others ones. I don't think toggle will work here.

nerik: `toggle` method name is a bit misleading, but if you specify a 2nd argument as I do here, it will not toggle but force the boolean value sent as the 2nd argument instead. So yes, it will work, and it's more concise.

nerik: Ref https://developer.mozilla.org/en/docs/Web/API/Element/classList

agnlez:
Didn't know about that hidden behaviour of toggle! You have changed my life _._ About dropdown styles.. I moved it, but I'm not sure about styling texts as the displayed one. It's too big in my opinion. But well, let's leave this to JC.

html/index.ejs
```diff
@@ -19,34 +19,80 @@
...
+            <li class="supply-chain-item">
+              <svg class="icon icon-supply icon-supply_chain_2"><use xlink:href="#icon-supply_chain_2"></use></svg>
+              <span class="description -bottom paragraph -center">Who trades <br/>in them?</span>
+            </li>
+            <li class="supply-chain-item">
+              <span class="description -top paragraph -center">Where are they consumed?</span>
+              <svg class="icon icon-supply icon-supply_chain_3"><use xlink:href="#icon-supply_chain_3"></use></svg>
+            </li>
+            <li class="supply-chain-item">
+              <svg class="icon icon-supply icon-supply-chain-4"><use xlink:href="#icon-supply_chain_4"></use></svg>
```
nerik: Please be consistent with icon-supply-chain-x and icon-supply_chain_x

styles/components/homepage/workflow-graphic.scss
```diff
@@ -0,0 +1,77 @@
...
+      display: flex;
+      position: relative;
+      flex-basis: 20%;
+      flex-direction: column;
+      align-items: center;
+      justify-content: center;
+
+      z-index: 1;
+
+      &:first-child > .icon-supply,
```
nerik: .icon-supply-chain-1

styles/components/homepage/workflow-graphic.scss
```diff
@@ -0,0 +1,77 @@
...
+      > .icon-supply-chain-4,
+      &:last-child > .icon-supply {
```
nerik: .icon-supply-chain-5
etc

nerik (changes requested):

nerik:
Ready to merge @andresgnlez ?

PR #61, opened 2016-10-24
homepage posts

agnlez:


scripts/pages/index.page.js
```diff
@@ -37,10 +39,32 @@ const _onSelect = function(value) {
   _setMap();
 };
 
+const _getPosts = () => {
+  const postList = document.querySelector('.js-posts-grid > .wrap');
+
+  d3_json('homepage/posts.json', (err, data) => {
```
nerik: Why didn't you use fetch/whatwg-fetch + JSON.parse ? This will add an unnecessary dependency to D3 in the index bundle.

nerik (changes requested):

nerik:
Please fix that JSON thing, and also rename the commit please. Otherwise looks good 👍

agnlez:
This is a `v1.0` of the dynamic grid. There are still some features to implement like choosing the position of vertical posts (left/right) but they are not critical and can be done in the next days.

PR #63, opened 2016-10-27
App navigation

agnlez:
nav
- Added toggle tool navigation in flows page.
- Unified both navigation bars (app and tool ones) into one.


styles/_helpers.scss
```diff
@@ -4,5 +4,5 @@
 }
 
 .is-hidden {
-  display: none;
+  display: none !important;
```
agnlez, line 7: I had to add the !important because for some reason, this class is being applied before the component styles. I don't know if you know how to fix this. I couldn't.

nerik: The reason is css selector specificity: because `.c-nav > .app-nav` is more specific (two classes instead of one) than `.is-hidden`, the display: flex rule will override the display: none.
Read: https://css-tricks.com/specifics-on-css-specificity/
!important is considered as bad practice, but in that case that is the least inelegant way I think. It makes sense because you want is-hidden to work in every possible case.

nerik: Curious to have @clementprdhomme's opinion on this.

scripts/containers/nav.container.js
```diff
@@ -18,7 +20,8 @@ const mapViewCallbacksToActions = () => ({
   onCommoditySelected: commodity => selectCommodity(commodity),
   onYearsSelected: years => selectYears(years),
   onQualSelected: qual => selectQual(qual),
-  onViewSelected: view => selectView(view)
+  onViewSelected: view => selectView(view),
+  onToggleAppMenu: (visibility) => toggleAppMenu(visibility)
```
nerik: Unneeded ()

scripts/components/nav.component.js
```diff
@@ -20,12 +22,35 @@ export default class {
...
+    this.toggleBtn = this.el.querySelector('.js-toggle-menu');
+  }
+
+  _setEventListeners() {
+    this.toggleBtn.addEventListener('click', () => this.onToggleMenu());
+  }
+
+  onToggleMenu() {
+    this.toggleBtn.classList.toggle('open');
+    this.callbacks.onToggleAppMenu();
```
nerik: Correct me if I'm wrong but I don't think the fact that the app menu is toggled is relevant to the rest of the app, nor is it necessary to have it as a sharable URL state.
Therefore you can keep the toggling logic entirely in `nav.component`, instead of going through the whole container/action/reducer/app store/container cycle.
This is important because I'm worried about our reducers and actions getting too big, so we should only use them when necessary.

scripts/containers/nav.container.js
```diff
@@ -9,7 +10,8 @@ const mapMethodsToState = (state) => ({
   selectCommodity: state.flows.selectedCommodity,
   selectYears: state.flows.selectedYears,
   selectQual: state.flows.selectedQual,
-  selectView: state.flows.selectedView
+  selectView: state.flows.selectedView,
+  setAppMenuVisibility: state.app.isAppMenuVisible
```
nerik: See comment about app store below.

scripts/reducers/app.reducer.js
```diff
@@ -18,6 +19,9 @@ export default function (state = initialState, action) {
   case actions.TOGGLE_MAP:
     return Object.assign({}, state, { isMapVisible: !state.isMapVisible });
 
+  case actions.TOGGLE_APP_MENU:
```
nerik: See comment about app store above.

nerik (changes requested):

nerik:
One other comment: when moving files around, always do it with git mv: this way you are sure that git will pick up the fact that you moved the file instead of removing and adding a new one, which makes comparing code between branches difficult.
Level 2: move files in a separated commit.

styles/_helpers.scss
clementprdhomme: @nerik is right here.
In the pattern we try to use, they call these kinds of rules "helpers": http://rscss.io/helpers.html

agnlez: Well, tbh I didn't add the `!important` because of that. I had no choice :/ Like @nerik I am not fan of using it unless you have no choice. Otherwise, thank you both for the feedback!

scripts/components/nav.component.js
agnlez: Yeah, I agree with you. I did it because I didn't want to change the way the app communicates unless it's necessary. What's the requirements to do it this way? Communication between components and the possibility of need for sharing or similar?

PR #65, opened 2016-10-28
API fixes

nerik:


nerik:
Pending https://github.com/sei-international/TRASE/issues/108

PR #66, opened 2016-10-28
Expand nodes

nerik:
1st objective is to be able to expand a single node (the latest selected)
This is needed for the search functionality.

We'll see later for multiple/cross column node expand


nerik:
expand

tiagojsag:
small detail: when you hover the button and it is close to the top, its top part is cut off.

nerik:
@tiagojsag https://github.com/Vizzuality/TRASE-frontend/issues/71#issue-186238452

nerik:
Will do a design pass when things settle down a bit

PR #73, opened 2016-10-31
Factsheets-actors page

hectoruch:
Add:

image


hectoruch:
Add:
- Grid lines

Delete: 
- Tick lines on chart

Update:
- Colors, font size, font family and other styles

image

package.json
```diff
@@ -48,6 +48,17 @@
   "dependencies": {
     "body-parser": "^1.15.2",
     "d3": "^4.2.2",
+    "d3-array": "^1.0.1",
```
nerik: There is no need to manually import all d3 modules, as the entry point in the `d3` module already has references to all d3 modules.

scripts/components/tables/area-tables.js
```diff
@@ -0,0 +1,33 @@
...
+
+  .selectAll('td')
+  .data(function(d){return d;})
+  .enter().append('td')
+  .style('border', '1px black solid')
+  .style('padding', '10px')
+  // .on("mouseover", function(){d3_select(this).style("background-color", "aliceblue")})
+  // .on("mouseout", function(){d3_select(this).style("background-color", "#fff1ae")})
+  .text(function(d){return d;})
+  .style('font-size', '12px');
```
nerik: While it is technically possible to build tables in d3 I strongly advise against it. Let's just use good old html tables for this.
We have a templating engine in the project that you can use, see example here for use with a table https://github.com/Vizzuality/TRASE-frontend/pull/77/commits/be5e9cab367e6a40c94498b579e3b16e6d0b5b2d

nerik (changes requested):

hectoruch:
Add styles for table:

image

package.json
agnlez: Agree. Please remove all those dependencies and use d3 modules already installed in the project. If you need any help with this, let us know.

html/factsheets.ejs
```diff
@@ -8,7 +8,110 @@
   <div class="l-factsheets js-factsheets">
     <%= htmlWebpackPlugin.options.nav %>
```
agnlez: You forgot about adding a `wrap` here

agnlez: Please, add `max-width` to the layout and use the variable already declared in `_settings` file.

scripts/components/graphs/area-stack.component.js
```diff
@@ -42,7 +42,7 @@ export default class {
 
     this.options = {
       width: this.el.clientWidth,
-      height: this.el.clientHeight
+      height: 380//this.el.clientHeight
```
agnlez, line 45: are we sure this is not going to change?

scripts/components/graphs/area-stack.component.js
```diff
@@ -143,6 +143,24 @@ export default class {
       g.append('g')
           .attr('class', 'axis axis--y')
           .call(d3_axis_left(y).ticks(5, 's'));
+
+      g.selectAll('.axis--y > .tick')
+        .select('line') //grab the tick line
+        .attr('x2', width)
+        .attr('stroke', '#34444c')
+        .attr('class', 'tick-line') //style with a custom class and CSS
+        .style('stroke-width', 1); //or style directly with attributes or inline styles
```
agnlez: Please, move styles to its styles file.

hectoruch:
Add icons on:
- Table 
  image
- Select area

image

styles/components/factsheets/area-select.scss
```diff
@@ -31,10 +31,22 @@
...
       .value-information {
+        display: flex;
+        align-items: center;
         margin-top: 7px;
         font-family: $font-family-1;
         font-size: $font-size-medium;
         letter-spacing: -1px;
+
+        .star-icon {
+          width: 16px;
```
agnlez: Icon sizes are setted in `icons` styles file. If you need to positionate them or similar do it here.

styles/components/factsheets/area-select.scss
```diff
@@ -31,10 +31,22 @@
...
+          margin-right: 2px;
+        }
+
+        .icon-check {
+          width: 16px;
+          margin-right: 4px;
```
agnlez, line 46: Same here.

package.json
```diff
@@ -1,65 +0,0 @@
-{
```
agnlez: Restore this! :P

scripts/components/graphs/area-stack.component.js
```diff
@@ -158,7 +158,8 @@ export default class {
       g.selectAll('.axis--x > .tick')
         .select('line') //grab the tick line
         .attr('class', 'tick-line') //style with a custom class and CSS
-        .attr('y2', 0);
+        .attr('y2', 0)
+        .style('stroke-width', 1); //or style directly with attributes or inline styles
```
agnlez: Move this to its CSS file, please.

html/factsheets.ejs
```diff
@@ -5,116 +5,118 @@
...
-          <ul class="js-dropdown-list dropdown-list">
-            <li class="js-dropdown-item dropdown-item" data-value="Soy">Soy</li>
-          </ul>
-        </div>
-        <div class="contain-information">
-          <div>
-            <span class="title-information">
-              ZERO DEFORESTATION COMMITMENT
+    <div class="l-factsheets">
+      <div class="content wrap">
```
agnlez: Don't use the `wrap` with something else, if you need another class or whatever, use another different element

styles/layouts/l-factsheets.scss
```diff
@@ -1,7 +1,9 @@
 .l-factsheets {
+  // max-width: $max-width-grid;
```
agnlez: Get rid of commented code.

Issue #75, opened 2016-10-31
Move scss imports to *.page.js files


nerik:
This way we allow webpack code splitting/chunks to properly work. See factsheets.scss for an example of what needs to be moved.
Additionally this allows us to benefit from webpack file aliasing.

agnlez:
In my opinion, pages shouldn't be treated as components because they aren't. If you want to move them because we already doing it, it's fine. You talked about importing `_base.scss` in GFW (actually it's not a good example...) but here you are importing `_settings`, so in the end it's the same but different file.

Anyway modularity has never gotten along with CSS structure.

nerik:
This is debatable, but the webpack argument stays.
Let's have a conversation with Clement about this after the deadline :)

PR #77, opened 2016-10-31
Place factsheets

nerik:
@hectoruch @andresgnlez I'm trying to set up a basic structure for fact sheets. This one's for places but it should be applicable to actors as well. Please have a look and let me know what you think.

Related:
https://github.com/Vizzuality/TRASE-frontend/issues/75
https://github.com/Vizzuality/TRASE-frontend/issues/76


scripts/pages/factsheet-place.page.js
```diff
@@ -0,0 +1,27 @@
+import 'whatwg-fetch';
+
+import 'styles/factsheet-place.scss';
+import 'styles/_base.scss';
+import 'styles/_texts.scss';
+import 'styles/components/shared/nav.scss';
+import 'styles/components/shared/_footer.scss';
```
agnlez, line 7: We have entry points for this, this way you only have to import one file here.

styles/factsheet-place.scss
```diff
@@ -0,0 +1,3 @@
+* {
+  color: red;
+}
```
agnlez, line 3: I guess this is testing... xD

styles/_texts.scss
```diff
@@ -1,3 +1,4 @@
+@import 'settings';
```
agnlez, line 1: Using an entry point you don't need to add the `settings` file everytime

styles/components/shared/_footer.scss
```diff
@@ -1,3 +1,4 @@
+@import '../../settings';
```
agnlez, line 1: Same here: https://github.com/Vizzuality/TRASE-frontend/pull/77/commits/b096668ac91c08bf1f14e56b9b471f2a494441f4#r86109916

agnlez:
Good job, I think it's far enough to start working on it! I just added some comments.

scripts/pages/factsheet-place.page.js
nerik: We have to avoid doing imports inside scss files, see https://github.com/Vizzuality/TRASE-frontend/issues/75
Importing scss files at JS component level is better because:
- it encapsulates style into logical components
- it helps aboiding the 'import base.scss' a million times like we had on fishing
- it allows webpack doing code splitting properly (ie put common css in the common bundle)

styles/_texts.scss
nerik: https://github.com/Vizzuality/TRASE-frontend/pull/77#discussion_r86112347

styles/factsheet-place.scss
nerik: Yes

styles/components/shared/_footer.scss
nerik: https://github.com/Vizzuality/TRASE-frontend/pull/77#discussion_r86112347

PR #80, opened 2016-11-02
Implemented app logo in flows page.

agnlez:
This PR implements the logo app in flows page. This way, the user is allowed to go back to the home if need it.

### DIFF
--- a/scripts/components/nav.component.js
+++ b/scripts/components/nav.component.js
@@ -41,6 +41,7 @@ export default class {
     this.AppNav = this.el.querySelector('.app-nav');
     this.FlowsNav = this.el.querySelector('.flows-nav');
     this.toggleBtn = this.el.querySelector('.js-toggle-menu');
+    this.logo = this.el.querySelector('.js-logo');
   }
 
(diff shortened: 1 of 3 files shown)


scripts/components/nav.component.js
```diff
@@ -49,14 +50,15 @@ export default class {
...
-    
+
     this.toggleBtn.classList.toggle('open');
     this.setAppMenuVisibility();
   }
 
   setAppMenuVisibility() {
     this.AppNav.classList.toggle('is-hidden', !this.state.visibilityAppMenu);
     this.FlowsNav.classList.toggle('is-hidden', this.state.visibilityAppMenu);
+    this.logo.classList.toggle('is-hidden', !this.state.visibilityAppMenu);
```
nerik, line 61: Runtime error here

nerik (changes requested):

scripts/components/nav.component.js
nerik: Seems I was drunk

PR #84, opened 2016-11-02
Implemented new map images

agnlez:
Also added logic to allow display country options base on the current commodity selection.


nerik:
Would be nice to select the 1st country in the list when changing commodity.

PR #94, opened 2016-11-03
Modal

agnlez:
- [x] Connect with app state.
- [x] API calls.


nerik:
Nice. I've added a bit of bootstrap code to get you started.

nerik:
Another remark: for map layers (i) buttons, I think they shouldn't trigger a modal but rather a small tooltip on hover (see https://basecamp.com/1756858/projects/12498794/todos/280717649#comment_473759929)

nerik:
Another remark: for map layers (i) buttons, I think they shouldn't trigger a modal but rather a small tooltip on hover (see https://basecamp.com/1756858/projects/12498794/todos/280717649#comment_473759929)

nerik:
Another remark: for map layers (i) buttons, I think they shouldn't trigger a modal but rather a small tooltip on hover (see https://basecamp.com/1756858/projects/12498794/todos/280717649#comment_473759929)

agnlez:
`http://localhost:8081/flows.html?story=1&state=eyJwYXJhbTEiOjEsInBhcmFtMiI6Mn0`

This is the URL format supported right now. Let me know if there is something has to be changed, please.

Right now, the app decodes the state if needed and add it to the defaults settings. I have added an error callback for API call: in case the call fails for any reason, the app will continue its flow.

image

And this is the modal appears by URL param. (Please, ignore the title, I'm not responsible of this)

PR #96, opened 2016-11-03
Remove unsupported opreator

tiagojsag:


nerik:
I'm not sure I understand the why here.
Furthermore, we have arrow functions all over the place, so why did you change only this one?

tiagojsag:
this particular lines make "npm run build" cry. I don't know why, tbh, but let me know if you can package for production without these changes - maybe it's a problem somewhere else

nerik:
Turns out this issue is only affecting the EJS webpack loader, not the code transpiled by babel.

PR #74, opened 2016-10-31
Map context layers

nerik:

PR #99, opened 2016-11-04
update trase contextual layer visualizations v0.1

aagm:


nerik:
Looks good! Silos don't appear below zoom level 5, is that normal?
(cherry-picked here https://github.com/Vizzuality/TRASE-frontend/pull/74)

nerik:
I added a forceZoom parameter that will change the zoom to the desired level when enabling a given context layer

PR #104, opened 2016-11-04
Component Table

hectoruch:
image


scripts/components/table/table.component.js
```diff
@@ -1,25 +1,36 @@
...
       });
   }
 
-  getData() {
-    return fetch('factsheets/table.json');
+  getData(value) {
+    if(value === 'top'){
+      return fetch('factsheets/tabletop.json');
+    }
+    if(value === 'municipallities'){
```
agnlez: You have a typo here. Is this working...?

hectoruch:
- [x] Improve dynamic logic for create tables.
- [ ] Tooltip -> Other branch.

scripts/components/table/table.component.js
```diff
@@ -4,27 +4,27 @@ import 'whatwg-fetch';
 export default class {
   constructor(value) {
     this.data = null;
-    this.getData()
+    this.getData(value)
       .then(response => response.json())
       .then((json) => {
         this.data = json;
         this.render(value);
```
agnlez: You are not using the `value` param here.

scripts/components/table/table.component.js
```diff
@@ -4,27 +4,27 @@ import 'whatwg-fetch';
...
     }
     if(value === 'municipalities'){
-      template = TableTemplate({actors: this.data});
+      return fetch('factsheets/table.json');
     }
+  }
+
+  render() {
+    var template;
+    var areamunicipalities = document.querySelector('.js-table');
```
agnlez: Change `var` to `const`

scripts/templates/table/table.ejs
```diff
@@ -12,8 +12,9 @@ var unit = [];
     <% } %>
   </tr>
 </thead>
+
 <tbody>
-  <% for(var i=0; i<title.length; i++) { %>
+  <% for(var i=0; i<actors['data'][title[0]].length; i++) { %>
```
agnlez: Use `let` instead of `var`

agnlez:
@hectoruch rebase this, please.

scripts/components/table/table.component.js
```diff
@@ -1,25 +1,30 @@
 import TableTemplate from 'ejs!templates/table/table.ejs';
-// import TableTopTemplate from 'ejs!templates/table/tableTop.ejs';
 import 'whatwg-fetch';
 
 export default class {
   constructor(value) {
     this.data = null;
-    this.getData()
+    this.getData(value)
```
agnlez: The idea is think about this component as an stupid one. It should receive data, organize it in a propery way and render it. That's all. Avoid make calls here, please.

scripts/components/table/table.component.js
```diff
@@ -1,25 +1,30 @@
...
 
-  getData() {
-    return fetch('factsheets/table.json');
+  getData(value) {
+    if(value === 'top'){
+      return fetch('factsheets/table2.json');
+    }
+    if(value === 'municipalities'){
+      return fetch('factsheets/table.json');
+    }
```
agnlez: This should be removed once we got API calls.

scripts/components/table/table.component.js
```diff
@@ -1,25 +1,30 @@
...
   }
 
-  render(value) {
-    var areamunicipalities = document.querySelector('.js-municipalities-table');
-    const template = TableTemplate({actors: this.data});
+  render() {
+    var template;
```
agnlez: Use `const` instead of `var`

scripts/components/table/table.component.js
```diff
@@ -1,25 +1,30 @@
...
+    var areamunicipalities = document.querySelector('.js-table');
```
agnlez: This element where the table is going to be render should be setted as a setting of the component.

scripts/pages/factsheet-place.page.js
```diff
@@ -5,24 +5,28 @@ import 'styles/_base.scss';
...
 fetch(`${API_URL}/v1/get_place_node_attributes?node_id=${nodeId}&country=BRAZIL&commodity=SOY`)
   .then(response => response.json())
   .then((data) => {
-    _build(data.data);
+    // _build(data.data);
   });
 
 const _build = data => {
-  console.log(data)
+  // console.log(data)
```
agnlez: Remove.

scripts/templates/table/table.ejs
```diff
@@ -1,21 +1,24 @@
+<%
+var title = [];
+var unit = [];
```
agnlez: Uhm... Not sure if I have understood this... Are you looping the data and setting variables in a template? Never do that. Templates are for render vars, the less logic, the better. If you need to organize the data before render it, use some method inside the component.

PR #109, opened 2016-11-07
Change contact us link

hectoruch:


agnlez:
@hectoruch probably you have to change the email again as the client say in this thread: https://basecamp.com/1756858/projects/12498794/messages/63691874#comment_474556836

Anyway, ask @tiagojsag before do something.

PR #110, opened 2016-11-07
Add tooltip component

hectoruch:


hectoruch:
image

scripts/components/map-layers.component.js
```diff
@@ -118,17 +119,30 @@ export default class {
...
   _onInfo(e) {
-    const target = e && e.currentTarget;
+    const target = e && e.currentTarget; //the information icon
+    const bounds = target.getBoundingClientRect(); //position of the icon
+    const top = bounds.top; // top
+    const left = bounds.left + 30; //left
     const uid = target.closest('.layer-item').getAttribute('data-layer-uid');
+    
+    var node = document.querySelector('.tooltip-layout');
+    node.style.display = 'block';
```
agnlez: Use `is-hidden` class instead of direct styles. The fewer direct styles the better.

scripts/components/map-layers.component.js
```diff
@@ -118,17 +119,30 @@ export default class {
...
+
+    node.style.top = `${topTooltip}px`;
+    node.style.left = `${left}px`;
+    node.innerHTML = `showing info of ${uid}`;
+  }
 
-    console.info(`showing info of ${uid}`);
+  _outInfo() {
+    var nodeOut = document.querySelector('.tooltip-layout');
+    nodeOut.style.display = 'none';
```
agnlez: Same here, `is-hidden` class.

styles/components/tooltip/_tooltip.scss
```diff
@@ -0,0 +1,30 @@
+  .tooltip-layout {
+  display: none;
+  position: absolute;
+  width: 300px;
```
agnlez: `width: 100%`

styles/layouts/l-flows.scss
```diff
@@ -1,5 +1,6 @@
 @import '../base';
 @import '../settings';
+@import '../components/tooltip/tooltip';
```
agnlez: `../components/shared/tooltip`

scripts/components/map-layers.component.js
```diff
@@ -118,17 +119,30 @@ export default class {
...
   // TODO: develop info function once is clear how it works
   _onInfo(e) {
-    const target = e && e.currentTarget;
+    const target = e && e.currentTarget; //the information icon
+    const bounds = target.getBoundingClientRect(); //position of the icon
+    const top = bounds.top; // top
+    const left = bounds.left + 30; //left
     const uid = target.closest('.layer-item').getAttribute('data-layer-uid');
+    
+    var node = document.querySelector('.tooltip-layout');
```
agnlez: Set this as property of the class `this.node` instead of look for it every time. Otherwise, I would choose another understandable name like `this.tooltip`.

styles/components/tooltip/_tooltip.scss
```diff
@@ -0,0 +1,30 @@
+  .tooltip-layout {
+  display: none;
```
agnlez: Don't set this here. Use CSS class based on incoming values instead.

scripts/components/map-layers.component.js
```diff
@@ -35,7 +35,8 @@ export default class {
 
   _setEventListeners() {
     this.infoBtns.forEach((infoBtn) => {
-      infoBtn.addEventListener('click', (e) => this._onInfo(e));
+      infoBtn.addEventListener('mouseover', (e) => this._onInfo(e));
+      infoBtn.addEventListener('mouseout', () => this._outInfo());
```
agnlez: Use better `mouseenter` and `mouseleave` (the event triggers the necessary times)

scripts/components/map-layers.component.js
```diff
@@ -118,17 +119,30 @@ export default class {
...
+    var topTooltip = top - (node.getBoundingClientRect().height/2) + 4;
+
+    node.style.top = `${topTooltip}px`;
+    node.style.left = `${left}px`;
+    node.innerHTML = `showing info of ${uid}`;
+  }
 
-    console.info(`showing info of ${uid}`);
+  _outInfo() {
+    var nodeOut = document.querySelector('.tooltip-layout');
```
agnlez: Use `const`

scripts/components/map-layers.component.js
```diff
@@ -118,17 +119,30 @@ export default class {
...
-    const target = e && e.currentTarget;
+    const target = e && e.currentTarget; //the information icon
+    const bounds = target.getBoundingClientRect(); //position of the icon
+    const top = bounds.top; // top
+    const left = bounds.left + 30; //left
     const uid = target.closest('.layer-item').getAttribute('data-layer-uid');
+    
+    var node = document.querySelector('.tooltip-layout');
+    node.style.display = 'block';
+    var topTooltip = top - (node.getBoundingClientRect().height/2) + 4;
```
agnlez: Use `const`

agnlez:
First all, rebase branch wit `develop`, please.

The tooltip is not vertically sided with its icon.

Actually, this is NOT a component, it doesn't have its own logic. I guess we can go ahead with this for this release, but this is not the idea of component. Thoughts, @nerik ?

nerik:
Maybe it's a component in the CSS sense, the definition of a component is quite loose anyways.

More importantly, I think it would be nice if you could help Hector set up the linters on his IDE if it's not already the case. If it is already, well ¯_(ツ)_/¯

PR #112, opened 2016-11-07
Add discard button

hectoruch:


hectoruch:
Add

image

tiagojsag:
@nerik can you take a look at my awesome js code, see if it isn't too magical. FYI the way I envision this feature, is more like a half reset - so all options on the top menu are not affected, just the node selections.

nerik:
👍

PR #113, opened 2016-11-07
Restore municipalities mapg

tiagojsag:


nerik:
Integrated those edits in https://github.com/Vizzuality/TRASE-frontend/pull/107

PR #116, opened 2016-11-08
fix table component

hectoruch:


styles/components/factsheets/area-table.scss
```diff
@@ -100,4 +92,8 @@
       }
     }
   }
+
+  &.-actors {
+    margin-top: 56px;
```
agnlez: Don't set position, margin, size properties here. Just common properties. Other ones go to layouts.

scripts/pages/factsheets.page.js
```diff
@@ -1,4 +1,8 @@
 import Dropdown from 'components/dropdown.component';
+import 'styles/_base.scss';
+import 'styles/_texts.scss';
+import 'styles/layouts/l-factsheets-place.scss';
+import 'styles/components/factsheets/area-table.scss';
```
agnlez: Remove this import: this file should be imported in its component. Once you call this component, it will import its own styles.

scripts/templates/table/table.ejs
```diff
@@ -1,22 +1,34 @@
...
+        <% } %>
+        </tr>
+      <% } %>
+    </tbody>
+  <% } else {%>
+    <tbody>
+      <% for(let i=0; i<actors.length; i++) { %>
+        <tr>
+          <td><%=i+1%>.<%= actors[i]['name'] %> <svg class="icon icon-check"><use xlink:href="#icon-outside-link"/></svg></td>
+          <td data-unit="<%= i == 0 ? '%' : ''  %>"><%= (actors[i]['value'] * 100).toFixed(2) %></td>
```
agnlez: `.toFixed(2)` <-- avoid logic in templates, please. Do this in the view, before render it.

scripts/components/table/table.component.js
```diff
@@ -1,25 +1,31 @@
...
+          this.render();
+        });
+    } else {
+      this.data = settings.data; //now a json next API
+      this.render();
+    }
   }
 
   getData() {
     return fetch('factsheets/table.json');
```
agnlez: You could just move the fetch sentence instead of creating a method for this.

PR #125, opened 2016-11-09
Fix dropdown select

hectoruch:


agnlez:
@nerik you added those styles for some reason, do you really need them? Those ones are showing selectors up badly, so we need to remove them.

nerik:
@andresgnlez Thx for the heads up. I added this bit because the top nav dropdowns in the tool can have very long titles, messing up the whole layout.
We need to keep this, can you please move them to class and add the class to the top nav dropdowns in the tool?

agnlez:
@hectoruch please, could you manage this? Thanks!

hectoruch:
Yes! @andresgnlez  @nerik

PR #127, opened 2016-11-09
Recolor by inds

tiagojsag:
Some elements on the "recolor by" are inds, so we need a bit more logic on that side to pass that info from the interface to the state container, and then to the query builder

I also injected that same state into the D3 rendering, so that we can have CSS classes that contain the qual/ind name, which should avoid colisions of CSS class names


nerik:
Awesome, muchas gracias 👍 
The only thing missing: coloring by ind-smallholders doesn't work because the API returns floating values while CSS classes expect ints. The other consequence is that you have an insane amount of links (possibly as much as there are values between 0 and 5, ie infinity)

tiagojsag:
https://basecamp.com/1756858/projects/12498794/todos/281419925#comment_475554317

tiagojsag:
just tested on the latest api, it is now working

PR #129, opened 2016-11-09
error message callbacks on factsheets pages

agnlez:
Also setted `wrapper-loading` layer properly to wrap whole code and avoid seeing load components and styles step b y step.

### DIFF
--- a/html/factsheet-actor.ejs
+++ b/html/factsheet-actor.ejs
@@ -80,7 +80,7 @@
 
           </div>
 
-          <div class="c-area-table ">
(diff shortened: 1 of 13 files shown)


agnlez:
NOTE: merge first #128 to remove https://github.com/Vizzuality/TRASE-frontend/pull/129/commits/e26ddf11a0f87b225e472928c59260260880549b of this PR.

agnlez:
This PR is pending on API response. Right now even for an unreal node id is returning OK (200) and it shouldn't. This makes uneasy manage error callbacks.

PR #131, opened 2016-11-10
Change styles for homepage

hectoruch:
image

### DIFF
--- a/html/index.ejs
+++ b/html/index.ejs
@@ -25,19 +25,19 @@
         <div class="c-workflow-graphic">
           <ul class="supply-chain-list">
             <li class="supply-chain-item">
-              <span class="description title -mono-font -center">Where are they produced?</span>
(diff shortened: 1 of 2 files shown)


html/index.ejs
```diff
@@ -22,33 +22,33 @@
...
-
         <div class="c-workflow-graphic">
           <ul class="supply-chain-list">
             <li class="supply-chain-item">
-              <span class="description paragraph -center">Where are they produced?</span>
+              <span class="description title -mono-font -center">Where are they produced?</span>
             </li>
             <li class="supply-chain-item">
-              <span class="description paragraph -center">Who trades <br/>in them?</span>
+              <span class="description title -mono-font -center">Who trades <br/>in them?</span>
```
agnlez: Please use the `description` class to make any change you need instead of using `title` and `-mono-font` classes.

PR #133, opened 2016-11-10
Added options to commodities selector in both factsheets pages

agnlez:


agnlez:
Not adding the `data-value` would set automatically the option as disabled one (with its styles)

PR #145, opened 2016-11-10
added html legends config

aagm:
we need to add the styles and the legend for the raster (waiting javier mail for that one)


nerik:
screen shot 2016-11-11 at 12 42 13

tiagojsag:
@nerik this breaks the styling on the sankey's search after rebasing with develop

PR #147, opened 2016-11-10
Multi node expand view

tiagojsag:


nerik:
This breaks a few things :D
I'll have a closer look tomorrow.

tiagojsag:
If this takes a long time to fix, leave it for later

PR #150, opened 2016-11-11
Add util for modify decimal numbers

hectoruch:


nerik:
It is partially a duplicate of https://github.com/Vizzuality/TRASE-frontend/blob/develop/scripts/utils/niceNumber.js

Can you merge the two? Always use `toLocaleString` for the end result because it properly shows the decimal separator depending on locale (, for US, . for Spain)

scripts/utils/changeNumber.js
```diff
@@ -0,0 +1,23 @@
...
+        numberChange = number.toFixed(10);
+      } else {
+        numberChange = 0;
+      }
+    }
+  } else{ // if is null
+    numberChange = 0;
+  }
+
+  return numberChange.toLocaleString();
```
agnlez, line 22: Why `String` ?

agnlez: Didn't notice about @nerik advice. Ignore this.

hectoruch:
please @andresgnlez review it.

PR #159, opened 2016-11-11
added loader wrapper to show nothing in slow connections

agnlez:


tiagojsag:
@andresgnlez this was supposed to be for all the pages, not just the fact sheets

PR #164, opened 2016-11-11
hasFlows

nerik:


nerik:
Do not merge until the flag has been implemented API side

PR #167, opened 2016-11-11
Fix forest 500 score and zero defor commitment

tiagojsag:


tiagojsag:
@andresgnlez please add the svg for the "no" icon

PR #180, opened 2016-11-28
Moved top table to a new top component

agnlez:


agnlez:
@hectoruch I have separated Top "table" to a new component... Things were becoming complicated. Take a look please.

hectoruch:
All works fine! @andresgnlez 👍

scripts/components/factsheets/top.component.js
```diff
@@ -0,0 +1,56 @@
...
+
+  _parseData() {
+    let baseUrlLink;
+
+    switch (this.targetLink) {
+    case 'place':
+      baseUrlLink = '/factsheet-place.html?nodeId=';
+      break;
+    case 'actor':
+      baseUrlLink = '/factsheet-actor.html?nodeId=';
```
nerik: I think you can make this simpler by simply using a template string (ie '/factsheet-${this.targetLink}.html?nodeId=')

scripts/pages/factsheet-place.page.js
```diff
@@ -26,25 +27,34 @@ const defaults = {
...
-    type: 'top'
-  });
+
+  new Top({
+    el: document.querySelector('.js-top-trader'),
+    data: data.top_traders,
+    targetLink: 'actor',
+    title: 'top traders',
+    unit: '%'
+  }).render(renderCallback);
```
nerik: I don't understand the need for a callback, as there is nothing async in render()

agnlez: There isn't yeah. But this way you allow the component to do something after call its render function and the component doesn't care about what is running, it just does it if it exists.

nerik: I still don't see the point. You do :
```
new Top (...)
new Top (...)
this.el.parentElement.classList.toggle (...)
```
it's functionally the same without the unnecessary tight coupling between the component and its parent.

scripts/utils/changeNumber.js
```diff
@@ -3,7 +3,7 @@ export default (number, type)  => {
 
   if(number != null){
     if(number > 0) {
-      if(type === 'top') {
+      if(type === 'percentage') {
```
nerik, line 6: Since type can only be undefined or 'percentage', can we replace this with an isPercentage boolean?

agnlez: I'll work on this in another branch, I want to make several changes here.

PR #183, opened 2016-11-29
Parse number function

agnlez:


scripts/utils/formatNumber.js
```diff
@@ -0,0 +1,19 @@
+export default (number, isPercentage)  => {
+  let parsedNumber;
+
+  if (number === null || number === 'undefined' || typeof number !== 'number') return 'N/A';
```
nerik: Beware, `undefined !== 'undefined'`
You need to do `number === undefined`

nerik: Can number be a string representing a number sometimes ? (ie `'42'`). In that case maybe we should do a parseFloat, then check that `typeof === 'number'`

scripts/utils/formatNumber.js
```diff
@@ -0,0 +1,19 @@
...
+
+  if (number === 0) return number;
+
+  if (isPercentage) {
+    parsedNumber = (number * 100).toFixed(1);
+  } else {
+    if (number > 0 && number < 1) {
+      parsedNumber = number.toFixed(4);
+    } else {
+      parsedNumber = number.toFixed(2);
```
nerik: On the sankey we are using 1 decimal, not 2. Are you sure the clients wants 2? If so we need to change it in `niceNumber` as well

scripts/utils/formatNumber.js
```diff
@@ -0,0 +1,19 @@
...
+    }
+  }
+
+  return parseFloat(parsedNumber).toLocaleString();
+};
```
nerik: FYI, instead of doing 

`parseFloat(number.toFixed(4)).toLocaleString()`

It's also possible to do:

`number.toLocaleString(undefined, {
  minimumFractionDigits: 0,
  maximumFractionDigits: 4
});
`

It's longer but you have more control over the number of decimals.

scripts/utils/formatNumber.js
tiagojsag: I recall a requests for 2 decimals on the fact sheets. Let's leave it as is, and we'll see if we get any extra requests from them.

tiagojsag:
if we are going to do this, we should do it for all values. see the three tables in  /factsheet-actor.html?nodeId=33

agnlez:
There are "two" types of tables and parse changes on every table. That's why in this table the format is not working. I'll check i out tomorrow.

PR #197, opened 2017-02-15
Feature/expand multiple nodes

pjosh:
Multiple node selection in the expanded view


nerik:
Awesome! This even works across columns!.
You just made SEI people happy :)

PR #198, opened 2017-02-16
Perfs master

nerik:


nerik:
Note: check that deploy correctly installed the Leaflet version specified in package.json

PR #199, opened 2017-02-20
Feature/select styles

pjosh:
* Layout adjust, basically margins and distance with elements.
* New style for the columns selectors.


pjosh:
The design will be changed again, so this changes probably don't have sense.

nerik:
That's a first step :)

PR #200, opened 2017-02-20
Initial API v2 support

tiagojsag:


scripts/utils/getURLFromParams.js
```diff
@@ -1,13 +1,42 @@
...
+      }, '');
       return `${prev}&${arrUrl}`;
     }
     return `${prev}&${current}=${params[current]}`;
-  }, `${API_URL}${endpoint}?`);
-};
+  }, `${API_V1_URL}${endpoint}?`);
+}
+
+export function getURLFromParams(endpointData, params) {
```
nerik: Why not passing strings instead of object endpointData? API_ENDPOINTS should be internal and only the key should be needed

nerik (dismissed):
The structure of this looks good, apart from a small encapsulation remark.
Note: endpoints are not working with provided ENV values.

scripts/utils/getURLFromParams.js
tiagojsag: I don't like string comparison, prefer to use constants or something in those lines.

nerik: Sure, put strings into constants.

nerik: But keep the actual data in it's area of concern.

nerik:
```
export const GET_ALL_NODES = 'GET_ALL_NODES';
const API_ENDPOINTS = {
   [GET_ALL_NODES]: { ... }
}
```

nerik:
@tiagojsag I fixed the map (sankey won't load for me)
There are a few things that are expected from the API that are currently mocked in the reducer, see https://github.com/Vizzuality/TRASE-frontend/pull/200/commits/4af271a51f2efd9d54558b2f0e3a7d761d8bb7d7#diff-9c54ed800a6f9264f7ab4bbe57148b40R56

PR #203, opened 2017-02-21
Feature/more flows colors

pjosh:
captura de pantalla 2017-02-21 a las 12 26 34 2

### DIFF
--- a/scripts/reducers/flows.reducer.js
+++ b/scripts/reducers/flows.reducer.js
@@ -175,15 +175,7 @@ export default function (state = {}, action) {
       links = mergeLinks(state.unmergedLinks);
     }
 
-    let mapNodeColors = [],
-      selectedNodeColors = [];
-    if (nodesColoredBySelection && nodesColoredBySelection.length !== 0) {
-      //TODO: finish this so that we can color the map too
-      const colorsFromLinks = getColorsFromLinks(links);
-      mapNodeColors = colorsFromLinks.geoColorMap;
-      selectedNodeColors = colorsFromLinks.colors;
-    }
-    newState = Object.assign({}, state, { links, nodesColoredBySelection, recolorGroups, mapNodeColors, selectedNodeColors });
+    newState = Object.assign({}, state, { links, nodesColoredBySelection, recolorGroups});
     break;
   }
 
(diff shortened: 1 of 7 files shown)


scripts/reducers/flows.reducer.js
```diff
@@ -175,12 +175,15 @@ export default function (state = {}, action) {
       links = mergeLinks(state.unmergedLinks);
     }
 
-    let mapNodeColors = [];
+    let mapNodeColors = [],
+      selectedNodeColors = [];
     if (nodesColoredBySelection && nodesColoredBySelection.length !== 0) {
       //TODO: finish this so that we can color the map too
-      mapNodeColors = getMapColorsFromLinks(links);
+      const colorsFromLinks = getColorsFromLinks(links);
```
pjosh: @nerik what do you think about take the selected node colors from all the links? Is there a better way?

nerik: You can directly use `recolorGroups` I think (collect the values from the dict using `Object.values` for example)
This means you don't need to loop through all links for that, in fact you probably don't even need to add property to the app state, just reuse recolorGroups.
As a side note, `getMapColorsFromLinks` and the `mapNodeColors` on the state are not used. My bad I should have clean that up.

nerik (changes requested):
The dynamic legend looks really good. It seems that the legend doesn't work anymore for other things than node selection though.

nerik:
@pjosh I did a few edits, please have a quick look at this commit: https://github.com/Vizzuality/TRASE-frontend/pull/203/commits/b46b088720d851fff19c929364bdf8af4b8dd2f5

As in a React application, we try to avoid as much as possible duplicating app state (state.flows.selectedRecolorBy) into component state (this.state.value).
This is a smell that often leads to problems (single source of truth principle).
So when you need to evaluate multiple parts of the app state (selectedRecolorBy and recolorGroups in that case) to update a component, you can use the _comparedValue / _returnedValue trick in the container. Consider the container as the part that prepares data from the app state to deliver it in the optimal form to the 'dumb' component.

Apart from that, at some point we might want to investigate at the action/reducer level why state.flows.recolorGroups is a huge array with mostly undefined values.

PR #205, opened 2017-02-21
Feature/map variables metadata endpoint

nerik:
- [x] mock endpoint
- [x] render folders
- [x] acceptable folder style (complete design TBD later)
- [x] load default variable selection at start


pjosh:
nice!
@nerik do you want a merge into develop although don't have the final design?

nerik:
I think so, @tiagojsag can you confirm please?

nerik:
Thanks

PR #207, opened 2017-02-22
Feature/basemap refacto

nerik:
This will allow switching basemaps, taking into account potential extra labels layer.


pjosh (approved):
Awesome!

PR #208, opened 2017-02-23
Feature/improve expand button

pjosh:
* Expand button tooltip with actions
* New logic for the reposition


nerik:
Noice !

nerik:
@pjosh loool, no, it's the scss linter (there's a rule about properties order). I have it as an Atom plugin, so I correct things sometimes when those red underline bother me 😛 
Admittedly we should do that a bit more seriously.

tiagojsag:
@pjosh my awesome git hooks are your friends ;)

PR #210, opened 2017-02-23
multiple map node selection

nerik:
Allows multiple node selection from map.
If a selected node is currently not visible (ie it's aggregated), sankey switches to expanded nodes mode, with all currently selected nodes expanded (aggregated or not)
When toggling off a node from the map, it's the same behaviour as if clicking a node on the sankey: sankey stays in expanded mode, with the toggled off node unselected but still visible.

### DIFF
--- a/scripts/actions/flows.actions.js
+++ b/scripts/actions/flows.actions.js
@@ -300,6 +300,19 @@ export function loadMapContextLayers() {
   };
 }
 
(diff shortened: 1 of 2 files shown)


nerik:
multiple-node

pjosh (approved):
Perfect!

PR #211, opened 2017-02-23
Feature/map legend scale

nerik:
The choropleth legend is broken, but since it's gonna need to be reworked anyway, I'd say we can merge that.


nerik:
scale

pjosh (approved):
nice!

PR #212, opened 2017-02-24
Feature/more basemaps

pjosh:
* Now, the map menu supports changing the basemap.
* New map menu design.

We need to set the final basemaps in the `BASEMAPS` dictionary.


nerik:
Awesome, this is perfect!

PR #215, opened 2017-03-01
get_contexts

nerik:
Renders the menus/dropdowns.
TODO:
- [x] replace mock JSON by real API
- [ ] there's an issue when loading state from a hash: app state get filled in at decoding the hash, then get_contexts is loaded but doesn't trigger update in nav view, because it's rewriting the same values. Fix is to move decoding of the hash after the loading of get_contexts 
- [x] actually render quant/quals/years dropdowns depending on context data
- [ ] when selecting country, and currently selected commodity is not available (or the opposite), sankey should not be visible anymore, prompting the user to select another commodity and/or country


tiagojsag:
#yolo-merge @nerik can you please review? This is a big one, so even after merged, we should make sure nothing too bad slipped through

PR #216, opened 2017-03-01
Feature/select polygons perfs

nerik:
Makes get_linked_geoids faster by not having a full cross lookup between polygons present in the map and those that should be linked.
It is still too slow for a very big number of polygons (ie Bunge).
The fact that we copy the geometries from the base layer to the linked polygons layer was probably an oversight (I did that for z-ordering - linked layers should always be on top otherwise outlines look 'eaten', and also thinking we'd need drop shadows etc later).
A future fix could be to just change the choropleth colors 'in place' in the base layer. We might want to also look for hiding the linked layer instead of removing it.

### DIFF
--- a/scripts/components/map.component.js
+++ b/scripts/components/map.component.js
@@ -23,6 +23,7 @@ export default class {
     });
 
     this.contextLayers = [];
+    this.polygonFeaturesDict = {};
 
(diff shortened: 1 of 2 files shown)


pjosh (approved):
nice!

PR #218, opened 2017-03-07
Feature/new map legend

pjosh:
New styles for the map legend


tiagojsag:
Seems like we're missing the actual values on the the legend. Is this meant for another PR?

pjosh:
done @tiagojsag !

tiagojsag:
image

Seems we have inconsistent state on initial load

tiagojsag:
also, isn't that left axis upside down?

pjosh:
oh damn! the initial state has always works, I think that I did some mistake merging with develop... Let me check!
You're right with the axis, thanks!

pjosh:
@tiagojsag exactly! merge mistake 😉 
Now, it should work good!

PR #233, opened 2017-03-16
Fix/recolour by menu

nerik:


scripts/components/sankey.d3layout.js
```diff
@@ -164,6 +168,7 @@ const sankeyLayout = function() {
       //   link.ty += layerOffsets[tLayerIndex];
       // }
     });
+    console.log(n)
```
tiagojsag: nope

nerik: fixed (and I ran the git hooks init script ;)

PR #236, opened 2017-03-16
fix/foundation-load-locally

pjosh:
Load the foundation grid locally.

**Note: we have differents ways to manage the styles on the landings. We should unify for mental health


nerik:
Thanks and yes, totally agree.

PR #237, opened 2017-03-16
Feature/basemaps

nerik:
Added staellite, street and topo basemaps.


pjosh (approved):
Great!! Only one more thing, what do you think about remove the Positron basemap? I remember that Juanca told me about to remove them, but I left it to try the basemap switch feature until the new ones.

PR #239, opened 2017-03-17
fix/remove-radio-style

pjosh:
Fix the radio X style
trase data VS trase data 1


nerik:
Still a bit off in Firefox:
screen shot 2017-03-17 at 10 48 03

PR #241, opened 2017-03-17
fix/download-data-params

pjosh:
Sending the right params to the download endpoint


scripts/components/data-content.component.js
```diff
@@ -85,28 +85,32 @@ export default class {
   }
 
   _downloadFile() {
-    window.open(this._getDownloadURL());
+    const downloadUrl = this._getDownloadURL();
+    window.open(downloadUrl);
   }
 
   _getDownloadURL() {
     const baseURL = getURLFromParams(GET_DATA_DOWNLOAD_FILE);
```
tiagojsag: this function accepts a second argument and builds the query for you automatically, so you can simplify this whole logic.
Also, country_id and commodity_id were removed on the api, use context_id instead (I just uploaded that to the api a few minutes ago)

pjosh:
done @tiagojsag ! thx for the feedback!

PR #244, opened 2017-03-17
fixes screenshot dimensions

sorodrigo:
This PR prevents the screenshots from being clipped or changing on different resolutions.


sorodrigo:
before


after

PR #245, opened 2017-03-17
Handle missing geometry files

tiagojsag:


tiagojsag:
This doesn't fix the whole issue, but should be a start

PR #249, opened 2017-03-17
feat dropdown-improvements

sorodrigo:
This PR removes selected option from the dropdown list and prevents action on unique child dropdowns.


scripts/components/dropdown.component.js
```diff
@@ -46,15 +46,25 @@ export default class {
...
+  _uniqueChild() {
+    return (this.list.querySelectorAll('li').length === 1);
+  }
+
   selectValue(value) {
     // TODO friday hack, this should not happen
     if (value === undefined) {
       value = 'none';
     }
+    if(this.current) this.current.classList.remove('is-hidden');
```
tiagojsag: code styling. check your IDE settings

styles/components/dropdown.scss
```diff
@@ -24,6 +24,14 @@
     }
   }
 
+  &.-unique-child {
+    pointer-events: none;
+    cursor: default;
+    .dropdown-title:after {
+      display: none;
```
tiagojsag: this is being overridden by the arrow mixin style.

scripts/components/dropdown.component.js
```diff
@@ -9,7 +9,7 @@ export default class {
     this.title = this.el.querySelector('.js-dropdown-title');
```
tiagojsag, line 8: This particular component is used everywhere. I found out the hard way that changes to it can spread caos. please be sure to test this carefully and on all places it's currently used

styles/components/dropdown.scss
sorodrigo: Are you sure? I think its working

tiagojsag:
Something still seams to be off here.

PR #257, opened 2017-03-20
Update texts on static pages

tiagojsag:


tiagojsag:
context 5 only has 3 column groups

PR #270, opened 2017-03-28
Homepage and legend style tweaks

juancarlosalonso:
Changed some text styles and margins of the homepage and the colors for the buckets of the legend.


nerik:
Perfect, thanks 👍

PR #272, opened 2017-03-30
Fix/choropleth fixes

nerik:


nerik:
screen shot 2017-03-30 at 11 53 47
screen shot 2017-03-30 at 12 06 03

nerik:
Note: with a bivariate choropleth, zero class (white) is only used when both variables are === 0

PR #276, opened 2017-04-05
allow explicit parameters state from url params

nerik:
This makes explicitly linking to a flow page node possible.
additionally it fixes various issues with the choropleth.


nerik:
@pjosh I added the links to the tool page from the the place and actor factsheets, I'll leave styling them to you :)

pjosh (commented):
Nice!!

PR #277, opened 2017-04-06
feature/mock-factsheets-api-calls

pjosh:
* A simple mocks management
* Factsheets mocks

### DIFF
--- a/public/mocks/v1_get_actor_node_attributes.json
+++ b/public/mocks/v1_get_actor_node_attributes.json
@@ -1,0 +1,696 @@
+{
+  "data": {
+    "summary": "Amaggi was the 5th largest exporter of soy in Brazil in 2015, accounting for 3,917,933 tons. This is a 2% increase vs the previous year. Amaggi sources from 1,258 municipalities, or 23% of the soy production municipalities. The main destination of the soy exported by Amaggi is Russia, accounting for 32% of the total.",
+    "column_name": "IMPORTER",
+    "forest_500": 4.0,
+    "node_name": "CARGILL",
+    "rtrs_member": "YES",
+    "supply_change": "YES",
+    "supply_change_link": "HTTP://SUPPLY-CHANGE.ORG/COMPANY/CARGILL",
+    "top_countries": {
+      "includedYears": [
+        2015,
+        2016
+      ],
+      "lines": [
+        {
+          "name": "CHINA",
+          "geoId":"BR1702109",
+          "type": "line",
+          "style": "line-pink",
+          "values": [
+            6392774.5,
+            6689568.95
+          ]
+        },
+        {
+          "name": "THAILAND",
+          "geoId":"BR1702109",
+          "type": "line",
+          "style": "line-pink",
+          "values": [
+            588071.79,
+            651347.46
+          ]
+        },
+        {
+          "name": "NETHERLANDS",
+          "geoId":"BR1702109",
+          "type": "line",
+          "style": "line-pink",
+          "values": [
+            451999.62,
+            346720.79
+          ]
+        },
+        {
+          "name": "SPAIN",
+          "geoId":"BR1702109",
+          "type": "line",
+          "style": "line-pink",
+          "values": [
+            311777.15,
+            331056.08
+          ]
+        },
+        {
+          "name": "INDONESIA",
+          "geoId":"BR1702109",
+          "type": "line",
+          "style": "line-pink",
+          "values": [
+            302305.63,
+            158144.7
+          ]
+        },
+        {
+          "name": "FRANCE",
+          "geoId":"BR1702109",
+          "type": "line",
+          "style": "line-pink",
+          "values": [
+            289828.7,
+            304808.66
+          ]
+        },
+        {
+          "name": "UNITED KINGDOM",
+          "geoId":"BR1702109",
+          "type": "line",
+          "style": "line-pink",
+          "values": [
+            277289.93,
+            359728.78
+          ]
+        },
+        {
+          "name": "INDIA",
+          "geoId":"BR1702109",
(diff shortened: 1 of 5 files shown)


public/mocks/v1_get_actor_node_attributes.json
```diff
@@ -0,0 +1,696 @@
...
+      "includedYears": [
+        2015,
+        2016
+      ],
+      "lines": [
+        {
+          "name": "CHINA",
+          "geoId":"BR1702109",
+          "type": "line",
+          "style": "line-pink",
```
nerik: Style info is not needed here.

scripts/utils/getURLFromParams.js
```diff
@@ -23,8 +23,8 @@ const API_ENDPOINTS = {
   [GET_FLOWS]: { api: 1, endpoint: '/get_flows' },
   [GET_MAP_BASE_DATA]: { api: 2, endpoint: '/get_map_base_data' },
   [GET_LINKED_GEO_IDS]: { api: 2, endpoint: '/get_linked_geoids' },
-  [GET_PLACE_FACTSHEET]: { api: 1, endpoint: '/get_place_node_attributes' },
-  [GET_ACTOR_FACTSHEET]: { api: 1, endpoint: '/get_actor_node_attributes' },
+  [GET_PLACE_FACTSHEET]: { api: 1, endpoint: '/get_place_node_attributes', mock: 'mocks/v1_get_place_node_attributes.json' },
```
nerik, line 26: 👍

nerik:
Please have a look at why factsheets break (ie what parts of the mocks are not formatted correctly), we shouldn't merge if it's broken.
Normally there shouldn't be too buch too change.

public/mocks/v1_get_actor_node_attributes.json
pjosh: Done @nerik !

pjosh:
@nerik breaks because of the grouping nodes! I'll fix it

nerik:
Awesome 👍

public/mocks/v1_get_place_node_attributes.json
```diff
@@ -0,0 +1,767 @@
...
+              51,
+              70,
+              6,
+              30
+            ]
+          }
+        ]
+      },
+      {
+        "name": "Deforestation",
```
agnessa, line 81: How can I know which indicators fall under the categories "Deforestation" and "Other environmental indicators"?

public/mocks/v1_get_place_node_attributes.json
```diff
@@ -0,0 +1,767 @@
...
+              0.0,
+              0.0,
+              null,
+              1840.0,
+              0.69,
+              2.0
+            ]
+          },
+          {
+            "name": "State Ranking",
```
agnessa, line 67: Not sure I understand what the "State Ranking" should be returning? This was not present in the old API I think.

public/mocks/v1_get_place_node_attributes.json
```diff
@@ -0,0 +1,767 @@
...
+        2011,
+        2012,
+        2013,
+        2014,
+        2015
+      ],
+      "lines": [
+        {
+          "name": "Soy related deforestation",
+          "type": "area",
```
agnessa, line 283: Where is this coming from?

public/mocks/v1_get_place_node_attributes.json
```diff
@@ -0,0 +1,767 @@
...
+            609500.0,
+            841427.0,
+            822126.0,
+            800000.0,
+            900536.0,
+            1000000.0
+          ]
+        },
+        {
+          "name": "State Average",
```
agnessa, line 342: Not sure what this is?

public/mocks/v1_get_place_node_attributes.json
nerik: No, it's new. It represents the ranking of the municipality for the given indicator within the state that the municipality belongs to.
For example: Querencia is 13th for Soy deforestation within the Mato Grosso state.

public/mocks/v1_get_place_node_attributes.json
nerik: It's static, but we considered the logic should be on the backend, because that's the pattern we followed for the Tool endpoints (old API). 
I think in the old API Clement has a workflow to insert data from a static JSON (like 'which indicators' belongs to which category') in the database.

public/mocks/v1_get_place_node_attributes.json
nerik: Same kind of answer as above, we consider that the graph type (line or area chart) is too tied to the meaning of it to have this on the front-end, but it's debatable.

public/mocks/v1_get_place_node_attributes.json
nerik: This are the values for all the municipalities of the state, averaged. So if you're looking at Querencia, average of all municipalities of the Mato Grosso state.

public/mocks/v1_get_place_node_attributes.json
```diff
@@ -0,0 +1,767 @@
...
+          90962.9242510086,
+          67557.6124615741,
+          60244.4898332252,
+          46734.840899474,
+          7299.83576324497,
+          347.889602841214
+        ],
+        [
+          233968.733919099,
+          0,
```
nerik, line 636: @agnessa This needs some further explanation: this is designed to be a 2D matrix of trading quantities between top traders or consumers (with an aggregated 'others' node), and top municipalities of the current state (with an aggregated 'others' node).
Until now we haven't implemented the 'other municipalities' part, which is why there are so many zeroes, but we need to change this to have the top municipalities.

PR #278, opened 2017-04-10
Feature/actor factsheets countries map

nerik:


nerik:
trase-fs-map

nerik:
@pjosh I implemented a basic countries map for actors factsheets.
There's no actual data shown in the map though, it should be done by:
- binding the data from the API with a choropleth css class (using the iso2 code on the map): https://github.com/Vizzuality/TRASE-frontend/pull/278/files#diff-63856bdcfa863c5a1cd657106b918e97R63
- rendering the tooltip with data from the API: https://github.com/Vizzuality/TRASE-frontend/pull/278/files#diff-63856bdcfa863c5a1cd657106b918e97R72

pjosh (commented):
it works perfect! 👏 👏

PR #279, opened 2017-04-11
Feature/factsheets overall improvements

pjosh:
* HTML with grid
* New styles for the overall blocks
* Adjust margins and sizes of the small maps
* Linker buttons


nerik:
👍

Issue #320, opened 2017-04-19
actor municipalities map


nerik:
- [x] load topoJson used in tool
- [x] work with Agniezska to implement real data (needs more than 10 municipalities)
- [x] map choropleth and legend

sorodrigo:
This is the same implementation as in the world map, so I can take this one.

PR #327, opened 2017-04-19
Feature/actor municip map

nerik:
screen shot 2017-04-19 at 18 13 42


nerik:
Fixes 1st item of #320

pjosh (commented):
works great!

Issue #314, opened 2017-04-19
Top 3 anchor buttons


nerik:
- 1st one should scroll to 1st section
- implement in actor page

Issue #302, opened 2017-04-19
Complete implementation of the new grid structure


nerik:
- for the two bottom diagrams on the place page
- for the actor page

PR #336, opened 2017-04-20
Feature/profile header improvements

pjosh:
* Profile actor header with the new design
* Anchor button with the scroll behaviour
* Put the rest of the elements inside the grid (collateral damages)

captura de pantalla 2017-04-20 a las 13 38 38 2


pjosh:
fixed #330 
fixed #314 
fixed #302

nerik:
Looks good !
Two small comments:
- could we get rid of the dropdown arrows on the actor page, since they are not doing anything yet?
- there are spacing issues between the sections in the actor page
Thanks !

pjosh:
Done @nerik !
I had not adjust the margins because we have to adjust all these modules individually, but I have already settled them more decent 😉

nerik:
Thanks!

Issue #287, opened 2017-04-19
Finalize tweets section


nerik:
- ask for exact Twitter accounts
- Uncomment existing code
- link to twitter

nerik:
Server code: https://github.com/Vizzuality/TRASE-social

PR #341, opened 2017-04-21
Feature/tweets

nerik:
- fixes posts slider (adds a prev button)

- does not fix images that don't fit the posts box ratio - this is not fixable on the frontend. @tiagojsag, I recommend the CMS has an option to not put an image so that we can simply use a placeholder with the trase logo on the frontend.

screen shot 2017-04-21 at 18 01 19

- add a slider for tweets
You need to update env vars and run the local tweets server to test this feature: https://github.com/Vizzuality/TRASE-social
@tiagojsag could you please have a look on how we could deploy this server?

screen shot 2017-04-21 at 18 03 03


nerik:
https://github.com/Vizzuality/TRASE-frontend/issues/287

tiagojsag:
If you need a server, you probably should use the content server app instead. Setting up a single app to serve tweets seems like a bit overkill to me, especially when we already have one CMS-like app up and running

Issue #323, opened 2017-04-19
actor scatter plot


nerik:
- work with Agniezska for the data format
- d3 scatter plot
- change/animate scatter plot when user changes x-axis

PR #343, opened 2017-04-25
Feature/scatterplot component

nerik:
https://github.com/Vizzuality/TRASE-frontend/issues/323


pjosh:
@nerik the code is ready to review!

public/mocks/v1_get_actor_node_attributes.json
```diff
@@ -595,84 +595,142 @@
...
         },
         {
           "name": "Loss of biodiversity",
+          "chart_name": "Loss of<br/>biodiversity",
           "unit": null
         },
         {
           "name": "Land-based emmisions",
-          "unit": null
+          "chart_name": "Land-based<br/>emmisions",
```
nerik: Typo: emissions

public/mocks/v1_get_actor_node_attributes.json
```diff
@@ -595,84 +595,142 @@
...
-      },
       "dimensions_x": [
         {
-          "name": "Land use",
-          "unit": null
-        },
-        {
           "name": "Territorial Deforestation",
-          "unit": null
+          "chart_name": "Territorial<br/>Deforestation",
```
nerik: I would remove `chart_name`. It is redundant with `name`. Line breaks are purely presentationbal and should be dealt with CSS.

scripts/components/profiles/scatterplot.component.js
```diff
@@ -0,0 +1,181 @@
...
+        );
+      }.bind(this))
+      .on('mouseout', function() {
+        this.hideTooltipCallback();
+      }.bind(this));
+    }
+  }
+
+  _renderXswitcher() {
+    const template = stringToHTML(ScatterplotSwitcherTemplate({data: this.xDimension}));
```
nerik: We need to stop using `stringToHTML`. We can just do:
```
this.switcherEl.innerHTML = ScatterplotSwitcherTemplate({data: this.xDimension});
```

(provided you created this.switcherEl from an existing DOM element before.)

scripts/components/profiles/scatterplot.component.js
```diff
@@ -0,0 +1,181 @@
...
+
+    const newData = this._getFormatedData(selectedTabKey);
+    let allXValues = newData.map(item => item.x);
+    let x = d3_scale_linear()
+      .range([0, this.width])
+      .domain(d3_extent([0, ...allXValues]));
+    const xAxis = d3_axis_bottom(x)
+      .ticks(8)
+      .tickSize(-this.height, 0)
+      .tickPadding(9)
```
nerik: No need to change but just a remark: even if the d3 standard is to do this, I personally much prefer using CSS for that kind of things (if it's not too complex to achieve)

scripts/components/profiles/scatterplot.component.js
```diff
@@ -0,0 +1,181 @@
...
+      switcher.classList.remove('selected');
+    });
+    selectedSwitch.classList.add('selected');
+
+    const newData = this._getFormatedData(selectedTabKey);
+    let allXValues = newData.map(item => item.x);
+    let x = d3_scale_linear()
+      .range([0, this.width])
+      .domain(d3_extent([0, ...allXValues]));
+    const xAxis = d3_axis_bottom(x)
```
nerik: xAxis is already instanciated above. Check that it's necessary, and in that case refactor to avoid repeating.

scripts/components/profiles/scatterplot.component.js
```diff
@@ -0,0 +1,181 @@
...
+        if (i === 0) {
+          return null;
+        }
+
+        return abbreviateNumber(value, 3);
+      });
+
+    this.circles
+      .data(newData)
+      .transition()
```
nerik: So in the end mapping between nodes is made automatically when they have the same nodeId right?

styles/components/shared/infowindow.scss
```diff
@@ -44,3 +44,54 @@
     }
   }
 }
+
+.c-profile-tooltip {
```
nerik: There's a lot of repetition with above `.c-sankey-tooltip` component. Try to see if merging both into a generic infowindow would not be too complicated.

nerik (commented):
One thing: we need to start enforcing updating the Changelog as we go (so we can have a clear view of what we done in the next meeting)

nerik:
I added a few code style comments, but otherwise looks really good :)

pjosh:
nice feedback! thanks @nerik !!

scripts/templates/profiles/tooltip.ejs
```diff
@@ -0,0 +1,7 @@
+<div class="js-infowindow-title title"><%= title %></div>
```
nerik: Can you move this file to templates/shared please? I'm going to use it for the Tool as well.

styles/components/shared/infowindow.scss
```diff
@@ -44,3 +44,54 @@
...
+        font-size: 16px;
+        letter-spacing: -0.8px;
+      }
+    }
+
+    &:not(:last-child) {
+      border-bottom: 1px solid #ebebeb;
+    }
+  }
+
```
nerik: Please run this file through the SCSS linter, it has many errors. (in particular color literals should be avoided)

public/mocks/v1_get_actor_node_attributes.json
pjosh: yes... I hate do that but in this case with CSS It's impossible to do the same as the design.
I'll check with Juanca to catch a solution.

scripts/components/profiles/scatterplot.component.js
pjosh: oh! I didn't know you can change the scale dynamically with `.scale(x)` ^^ nice

nerik:
Awesome, thanks. Could you please update the Changelog, then we're good to go 👍

nerik:
Also:
screen shot 2017-04-28 at 13 24 20

Issue #344, opened 2017-04-26
hover effect on 3 top links/buttons


nerik:
Because it's not so obvious they are clickable

Issue #286, opened 2017-04-19
Separated entry points to tool from general menu


nerik:
Tool should be a dropdown menu with:
- supply chain actors: same as current link
- production regions: tool page with map expanded (use explicit URL with ?isMapVisible=true as in profile pages buttons)

nerik:
https://basecamp.com/1756858/projects/12498794/todos/298634092

nerik:
Same idea as here but for the Tool entry:
![screen shot 2017-04-26 at 17 12 56](https://cloud.githubusercontent.com/assets/1583415/25441851/ab60b8de-2aa3-11e7-80f2-7f9e5d5d1e75.png)

PR #345, opened 2017-04-26
Feat/tool nav submenu

sorodrigo:
- From #286 : divides the tool nav bar link into a submenu with 2 entries: Flow and Flow w/ map expanded.

**Bonus**:
Addresses the #344 missing hover effect on link buttons.


pjosh (commented):
cool! 👏

PR #357, opened 2017-04-28
Fix: long names municipality/actor profile

sorodrigo:
- [x] removes invisible search icon that was occupying space inside `link-button`.
- [x] formats texts that contain an `'s` in the end to prevent in the case where the name ends with `s`.
- [x] fixes text breaking in `link-button` when the name is too long.
- [x] fixes broken text in chords diagram when name is too long.


pjosh (commented):
👍 👍

PR #358, opened 2017-04-28
Feat: actors world/municipality map

sorodrigo:
- Mocks actors buckets.
    - [x] world-map
    - [x] municipalities-map
- Reads buckets from data and sets the choropleth classes accordingly.
    - [x] world-map
    - [x] municipalities-map
- Reads the country's values from data and sets the infowindow text accordingly.
    - [x] world-map
    - [x] municipalities-map
- Implements real data w/ more than 10 countries.
    - [ ] world-map
    - [ ] municipalities-map


nerik:
Don't forget the changelog please !

sorodrigo:
Ready for review 😁

pjosh:
The trade value on the tooltip should have units & title
trase actor profile VS 2 factsheets 2 actor trase v 1 5 zeplin

PR #362, opened 2017-05-03
fail-safe links loading, fixes #346

### LINKED ISSUE #346: Direct links from actor profile are broken

nerik:
This will reload links endpoint in the tool, without any parameters, when for some reason the API returns an empty list when trying to get links.

Test scenario: look for "Geym" in profiles, then click on the link that leads to the tool. It should load the links with the given node id, fail, and reload the links without parameters.


pjosh (commented):
works great!

PR #368, opened 2017-05-03
Feature/map tooltips

nerik:
When map is expanded, show info related to node as a tooltip directly on the map instead of below sankey.

trase-map-tooltips


pjosh (commented):
nice!! 👏 👏

Issue #366, opened 2017-05-03
Collapsable map dimensions groups


nerik:
- [x] dimension selected indicator

simaob:
Merged and deployed to staging. Closing.

PR #371, opened 2017-05-04
Feature/new map dimensions selector

nerik:


nerik:
trase-dimensions

nerik:
This :
- adds a new feature: map dimensions groups are now collapsible. This is kept in the URL state
- makes the map dimensions selection simpler, with a single radio column

Internally, in the store selected dimensions are now stored as a simple array of ids, rather than an object containing vertical/horizontal properties that don't apply anymore. Additionally CSS has been simplified and reorganized so that each component handles its own CSS (map dimensions, context and basemap)

nerik:
https://github.com/Vizzuality/TRASE-frontend/issues/366

pjosh (commented):
wow! 👏

PR #374, opened 2017-05-05
Feat: choro map legend

sorodrigo:
- New choro-legend component.
- Profile map legends.
- Changes profile map choropleth's scheme to match the one in the tool.
- Includes `n-a` style in the profile map choropleth.


pjosh (commented):
👌

PR #375, opened 2017-05-08
Charts + Choropleths tune-ups

sorodrigo:
- The dissappearing of the legend, when switching the municipalities source. this was due to the legend always rendering in the countries `.legend-container` div instead of on the municipalities one.
- Changes the scatterplot component to do a comparison based on the name of the node, because same actors have different nodeid's when they are both exporters and importers.


nerik:
Noice 👍

PR #376, opened 2017-05-08
Fix: charts polishing

sorodrigo:
- fixes tick padding of the xAxis in Line chart.
- Limits the xAxis ticks in the Line chart to 6.
- Displays months in the xAxis if the number of years is <= 2.
- Fixes order of the lines to display the highest most recent values first.


sorodrigo:
**Before:**
screen shot 2017-05-08 at 17 37 03



**After:**
screen shot 2017-05-08 at 17 32 56

pjosh (commented):
👍

PR #380, opened 2017-05-09
Fix: hide null dots scatterplot

sorodrigo:
- In the scatterplot when dots were null they appeared in the `x = 0` position. Now they're hidden when they don't have data.


pjosh (commented):
yeah!

PR #384, opened 2017-05-10
Fix: search bar styles

sorodrigo:
- Fixes z-index of search bar in tool that makes it appear behind nav.

[BONUS]
 - Changes `innerHTML` to `textContent` inside actors page.


nerik:
Thanks 👍

PR #403, opened 2017-05-23
prevent states and biomes from appearing as profile pages links

nerik:
States and biomes nodes are filtered out from both profile search and node links in tool.

@tiagojsag eventually this should come from the API, please check https://github.com/sei-international/TRASE/issues/269


tiagojsag:
I'll check if the API change is quick, so we can skip this hack-around

tiagojsag:
If possible, merge https://github.com/Vizzuality/TRASE-frontend/pull/419 instead

PR #404, opened 2017-05-23
fixed place profile global map

nerik:
Current country highlight was missing in place profile.
This fixes it by comparing with country name instead of iso code, which is missing from the world topo json, and I'm to lazy to regenerate one.


tiagojsag:
@nerik found a bug, so he'll look into it

nerik:
Needs https://github.com/Vizzuality/trase-api/issues/19.
See https://github.com/Vizzuality/TRASE-frontend/issues/414

PR #408, opened 2017-05-23
fixed switch column on search

nerik:
Fixes an issue that prevented a column switch when for example, looking at municipalities and searching for a state.


nerik:
https://basecamp.com/1756858/projects/12498794/todos/305530858

PR #416, opened 2017-05-24
use default color when one of the indicators is zero

nerik:
Here's how it looks with the following rule:
A polygon renders to 'default' (gray) when *one of* the dimensions is equal to zero

screen shot 2017-06-07 at 19 19 12

Compared to the former rule:
A polygon renders to 'zero' (white) when *both* dimensions are equal to zero
screen shot 2017-05-24 at 19 10 00


nerik:
Do not merge, currently discussing here https://basecamp.com/1756858/projects/12498794/todos/305532456

PR #418, opened 2017-05-25
Prevent browser from endlessly looping on API v1 404

tiagojsag:


nerik:
Looks good. 
A bit of context here would have been nice too. When do we have 404s? Do we know why?

tiagojsag:
the question is not so much "when" or "why", it's more like "if we do, crashing the browser/ freezing the computer is a bad error handling policy :P

nerik:
Makes sense 🤣

PR #424, opened 2017-05-26
Fully rename factsheets to profiles

tiagojsag:
https://basecamp.com/1756858/projects/12498794/todos/308418791


nerik:
Thanks, this was long overdue 👍 
I wouldn't merge that before we set up HTTP redirections for the old URLs on the production server.

tiagojsag:
we can only set redirections once this is merged and live, otherwise we break the feature entirely

nerik:
Yup :D
What I meant was that we should have this _ready_ before we merge.

PR #430, opened 2017-06-13
Restore and improve divisor logic

tiagojsag:
- restores the divisor logic - fixes the issue on https://basecamp.com/1756858/projects/12498794/todos/312148527
- Makes sure the intervals don't overlap, start at 0 and end at 100
- Uses grey color on "unknown" values

### DIFF
--- a/scripts/components/tool/sankey.component.js
+++ b/scripts/components/tool/sankey.component.js
@@ -137,8 +137,12 @@ export default class {
       return classPath;
     }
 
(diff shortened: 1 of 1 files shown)


nerik:
Looks like there are still some issues there
screen shot 2017-06-14 at 12 53 12
screen shot 2017-06-14 at 12 53 19

nerik:
Looks good with the API deployed 👍

PR #431, opened 2017-06-13
Fix/node cards duplicates

nerik:
Avoids rendering node cards and tooltips twice for the same dimension.
https://basecamp.com/1756858/projects/12498794/todos/312319406

Additionally, avoids rendering the unit in cards and tooltips when it's 'Unitless'.


nerik:
@tiagojsag https://github.com/sei-international/TRASE/issues/272

nerik:
Also see https://github.com/sei-international/TRASE/issues/271

tiagojsag:
Pending feedback in the basecamp thread linked above

PR #432, opened 2017-06-14
Fix issue where resize by would only load from url if recolor by was set

tiagojsag:
because the greatest bugs come from the smallest typos


nerik:
❤️

PR #433, opened 2017-06-14
Add partner links to logos on footer and about page

tiagojsag:


styles/components/shared/_footer.scss
```diff
@@ -65,19 +65,19 @@
             }
           }
 
-          &.sei-logo > img {
+          &.sei-logo > a > img {
```
nerik, line 68: Uou might as well remove the direct descendant rule altogether, ie 
```
&.sei-logo img
```

PR #434, opened 2017-06-14
Feature/mobile disclaimer

nerik:
screen shot 2017-06-14 at 18 24 09
screen shot 2017-06-14 at 18 24 17


tiagojsag:
image

And let's remove the "yet", as it makes it sound like we have plans to go mobile.

PR #436, opened 2017-06-16
Fixed map dimensions display in node cards

nerik:
- fixed an issue that prevented currently selected map dimensions values to ever appear on clicked/ "pinned" node cards
- fixed an issue that prevented currently selected map dimensions values to be updated when changing map dimensions

screen shot 2017-06-16 at 11 24 29


tiagojsag:
seems like we only have thousand separators on some dimensions, not all. Assuming this is a frontend issue, can you fix it, @nerik ? thanks

nerik:
I'm currently having a look at this as well as https://basecamp.com/1756858/projects/12498794/todos/298634250, on another branch

PR #439, opened 2017-06-30
Fix issue with selecting inexisting geoids onmap

tiagojsag:
Issue happens when you select a country node (4th column on sankey). Because they have a geoId, the current map tries to highlight their polygon on the map, and crashes


tiagojsag:
see also https://github.com/Vizzuality/trase-api/commit/c6b67d29e1c4ccd163d8843a9583ab29020b1e67

PR #445, opened 2017-07-19
Maintain cloropeth on map when non-geo column is changed

tiagojsag:
https://basecamp.com/1756858/projects/12498794/todos/315260842


nerik:
👍

PR #460, opened 2017-08-03
show dark logos in home

nerik:
trase-header


tiagojsag:
if you scroll down, and then all the way back up, the menu bar seems to remain white. Can we have it change back to red in that scenario?

nerik:
I couldn't reproduce in Chrome, Safari, FF
Do you have more details?

PR #462, opened 2017-08-23
Clear node selection button on node bar

nerik:
trase-clear


nerik:
~~[ ] replace "sticky" node menu with a single expand button behaving like this one~~

nerik:
@tiagojsag ready to review again

PR #468, opened 2017-09-05
deploy script

nerik:


bin/deploy.sh
```diff
@@ -0,0 +1,24 @@
...
+else
+  SERVER_URL='staging.trase.earth'
+fi
+
+echo "DEPLOYING:  $ENV_TYPE ($SERVER_URL)"
+
+read -p "Press enter to continue"
+
+cp .env .env.tmp
+cp .env.$ENV_TYPE .env
```
tiagojsag: https://www.cyberciti.biz/faq/unix-linux-test-existence-of-file-in-bash/

bin/deploy.sh
```diff
@@ -0,0 +1,24 @@
...
+
+npm run webpack:prod
+
+cp .env.tmp .env
+rm .env.tmp
+
+scp -pr ./dist/* ubuntu@$SERVER_URL:/var/www/html/
```
tiagojsag: you should restart nginx varnish and apache after this. see the command I left you on slack

package.json
```diff
@@ -12,7 +12,9 @@
     "eslint": "npm run lint-js",
     "lint-js": "eslint -c .eslintrc.js --ext .jsx,.js scripts",
     "lint-sass": "sass-lint -c .sass-lint.yml -v -q 'styles/**/*.scss'",
-    "postinstall": "npm run build"
+    "postinstall": "npm run build",
+    "deploy:staging": "./bin/deploy.sh staging",
```
tiagojsag, line 16: can't we do a single command, that accepts produtcion/staging as an argument?

bin/deploy.sh
tiagojsag: does this clean /dist before it starts?

nerik: > you should restart nginx varnish and apache after this. see the command I left you on slack

Even if we are just touching static files?

package.json
nerik: Actually I thought wrapping this would prevent running `./bin/deploy.sh monkey`. Since we only have 2 envs...

PR #474, opened 2017-10-03
Account for API change in var name on actor profile; minor validation…

tiagojsag:
…s to improve resilience; load companies chart label from API data


tiagojsag:
requires https://github.com/Vizzuality/trase-api/pull/43

nerik:
👍

PR #476, opened 2017-10-04
Progressively load more nodes on complete view

tiagojsag:


tiagojsag:
We don't want to currently use this solution, but might be worth revisiting in the future

PR #481, opened 2017-10-07
Default all data downloads as pivot

tiagojsag:


nerik (commented):
You had a console.log hanging around, please be careful with linter.

nerik:
LGTM

PR #482, opened 2017-10-10
Analytics review

tiagojsag:
Reviewed the GA tracking implementation
- Added new events for data portal
- Moved tracking logic from reducers and into dedicated middleware (could be DRYer, but that's for another day)
- Retested the previous events and fixed the broken ones
- Added new event to track context changes in sankey

Full event reference: https://docs.google.com/spreadsheets/d/11Pufihk7VuxvvHWs5CxKNIklazEcKsu7lRPL50InFg0/edit#gid=0

### DIFF
--- a/config/webpack.js
+++ b/config/webpack.js
@@ -108,6 +108,7 @@ const config = {
     root: process.cwd(),
     alias: {
       actions: 'scripts/actions',
+      analytics: 'scripts/analytics',
       reducers: 'scripts/reducers',
       templates: 'scripts/templates',
       style: 'styles',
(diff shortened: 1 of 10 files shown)


nerik:
DRYed the middleware code.
Maybe not fix for now but worth raising a flag: this generates a lot of noise GA events. For instance, a context switch will generate two other events that are not meaningful:
```
{hitType: "event", eventCategory: "Sankey", eventAction: "Select contextual layers", eventLabel: "brazil_biomes"}
{hitType: "event", eventCategory: "Sankey", eventAction: "Update node selection", eventLabel: ""}
```

tiagojsag:
yeah, there are also many actions generating calls to the "node selection" event. we'll have to take care of that at some point, but not the goal right now

PR #483, opened 2017-10-10
Static page content update

tiagojsag:
https://basecamp.com/1756858/projects/12498794/todos/326872743


tiagojsag:
Includes a style change on the <a> as requested per JC

html/data-methods.ejs
```diff
@@ -4,216 +4,168 @@
...
+          programmes of work on beef in Argentina, Brazil and Paraguay; palm oil in Indonesia and Colombia; and coffee
+          in Colombia. Additional countries and commodities will be added in subsequent years based on capacity and
+          funding opportunities.
+        </p>
+        <p class="paragraph">
+          While the modelling framework and decision logic that underpin SEI-PCS are transferable between different
+          commodities and countries, each version of SEI-PCS is highly tailored to the specific context, requiring
+          detailed knowledge of national production, logistics, taxation and other commodity-country-specific data. To
+          learn about plans for SEI-PCS and Trase expansion to other countries and commodities please <a
+                  href="/#newsletter" target="_blank">sign up</a> for our newsletter.
```
nerik: This leads to nothing as the inactive newsletter section of the homepage has been removed.

styles/layouts/l-FAQ.scss
```diff
@@ -31,6 +31,13 @@
       max-width: $content-max-width;
       margin: 0 0 0 $anchor-list-max-width;
 
+      a {
```
nerik: This block is duplicated 4 times. You could add for example a `content-link` class to those links and add the style to _texts.scss, or add a class to the body (`.content-page a`)  and add the style to _texts.scss

nerik (changes requested):
See comments.

PR #484, opened 2017-10-10
Add subscribe to newsletter on the homepage

tiagojsag:


scripts/pages/index.page.js
```diff
@@ -105,6 +106,38 @@ const scrollIntro = () => {
   }
 };
 
+const newsletterSubscribe = (e) => {
+  e.preventDefault();
+  const form = document.querySelector('.c-newsletter-form');
+
+  if (form.checkValidity() === false) {
+    return;
```
nerik, line 114: Some user feedback would be nice (reuse stuff used in server response handling)

nerik (changes requested):
See comment about user feedback.

scripts/pages/index.page.js
tiagojsag: browser handles that automatically, highlights the field in red.

PR #486, opened 2017-10-15
Change get_linked_geo_ids response handling to account for new respon…

tiagojsag:
…se format

Needs https://github.com/Vizzuality/trase-api/pull/56


tiagojsag:
@nerik we should review and merge this asap

PR #493, opened 2017-10-21
Replace get_nodes with get_node_attributes

tiagojsag:
Requires https://github.com/Vizzuality/trase-api/pull/61

The new call drops data on empty nodes, and also does not load layer metadata anymore (we have a different endpoint for that anyways)


tiagojsag:
As per my conversation with @agnessa we're freezing this PR for a while, until we are confident we can handle another moving part in the project

PR #495, opened 2017-10-24
Disable host checking by webpack server

tiagojsag:
In case we need to test this in a VM.


nerik:
I'm not familiar with this and the why of this, could you elaborate please? The <a href="https://webpack.js.org/configuration/dev-server/#devserver-disablehostcheck">webpack doc</a> says it's not recommended.

tiagojsag:
if you try to access the content served by webpack server from a different machine (VM using IE/Edge, for example), the webpack server will give you an error, basically because you are not using localhost/0.0.0.0

This disables this protection

nerik:
I'm not sure I get all the implications of this, but maybe read this first before merging:
https://medium.com/webpack/webpack-dev-server-middleware-security-issues-1489d950874a

tiagojsag:
Interesting read, and definitely worth keeping in mind, but doesn’t apply here, because we are not using the webpack server in production, we just use it locally.

nerik:
The webpack dev server is not intended for production. What that post says from my understanding is that a malicious website could make requests to a webpack dev server running locally (thus potentially exposing dev env vars, data served from a dev/local API, etc).

tiagojsag:
I’m 95% sure this only opens up a security issue on the machine running the webpack dev server, which is none in our case. But, just to be safe, maybe @agnessa can take a look

agnessa:
I got a headache from reading about this but I think Erik is right, it opens up an issue on localhost. The solution to the original issue of testing in VM would be to somehow whitelist the alternative hosts we want to allow rather than turn this off. To be fair, this possibly affects many other services which we run on our local machines, no need to search far and wide: https://twitter.com/homakov/status/839066663437787136

PR #496, opened 2017-10-25
Download profile as pdf

nerik:
Adds a button on the top right of the profile pages to download a printable, PDF version of the profile page, using an external endpoint.


tiagojsag:
I see 2 problems here:
- We need to check with the RW people if this service will planned to remain available as-is
- Pages that take a while to load (Bunge profile page, cold cache) are rendered with the loading spinner, instead of having the service wait for the ajax calls to complete.

Overall, I'd be happier if we could somehow copy over that service into Trase's server, and tweak it as needed. Might also be worth involving @agnessa in the discussion, as maybe "there's a ~~app~~ gem for that"

PR #498, opened 2018-02-08
Add reset sankey button to no flows error message

tiagojsag:


scripts/actions/tool.actions.js
```diff
@@ -44,6 +44,65 @@ export function resetState(refilter = true) {
...
   };
 }
+
+// Resets sankey's params that may lead to no flows being returned from the API
+export function resetSankey() {
+  return (dispatch, getState) => {
+    const state = getState();
+    const context = state.tool.contexts.find(context => context.id === state.tool.selectedContextId);
+    const defaultColumns = state.tool.columns.filter(column => column.isDefault);
+    const defaultResizeBy = context.resizeBy.find(resizeBy => resizeBy.isDefault);
```
sorodrigo: this is an unsafe access. I would add a fallback `context && context.resizeBy`

scripts/components/tool/tool-content.component.js
```diff
@@ -1,21 +1,27 @@
-
 export default class {
 
   onCreated() {
     this._setVars();
+
+    this.onSankeyReset = this._resetSankey.bind(this);
```
sorodrigo, line 6: It feels weird to bind in `onCreated` instead that in the `constructor`.

scripts/components/tool/tool-content.component.js
```diff
@@ -36,5 +42,11 @@ export default class {
 
   toggleError(hasError) {
     this.sankeyError.classList.toggle('is-hidden', !hasError);
+
+    if (hasError) {
```
sorodrigo: I think we can add the event listener on constructor, doesn't make sense to toggle add/remove depending on `hasError`

scripts/actions/tool.actions.js
tiagojsag: true, but handling a case where a context is missing would be a totally different beast, so I'm adding that validation, but not really handling the case where context is missing

scripts/components/tool/tool-content.component.js
tiagojsag: We're using the same pattern in other places, I'm just reusing it

scripts/components/tool/tool-content.component.js
tiagojsag: moved

